PR #29, opened 2021-06-21
Update dependencies

smoelius:


danielhenrymantilla:
I notice there is no `Cargo.lock` tracked in this repo; I recommend that it be done, so as to allow people to keep installing and using the binaries through the `--locked` flag on `cargo install`, should some dependency incorrectly introduce a breaking change within a semver-compatible version change (which, whilst rare, does happen).

And, in order to guarantee the `Cargo.lock` file is in sync with the `Cargo.toml`, I recommend that a
`cargo update -w --locked` step be added to CI.

smoelius:
Thanks for this suggestion, @danielhenrymantilla. I'm going to merge this PR and open an issue so that I can think about it some more.

Issue #30, opened 2021-06-21
`Cargo.lock` file?


smoelius:
In #29, @danielhenrymantilla wrote:
> I notice there is no `Cargo.lock` tracked in this repo; I recommend that it be done, so as to allow people to keep installing and using the binaries through the `--locked` flag on `cargo install`, should some dependency incorrectly introduce a breaking change within a semver-compatible version change (which, whilst rare, does happen).
> 
> And, in order to guarantee the `Cargo.lock` file is in sync with the different `Cargo.toml` files, I recommend that a
> `cargo update -w --locked` step be added to CI 🙂

PR #34, opened 2021-06-22
Check lockfiles

smoelius:
Addresses #30


scripts/update_lockfiles.sh
```diff
@@ -0,0 +1,13 @@
...
+set -euo pipefail
+
+if [[ $# -ne 0 ]]; then
+    echo "$0: expect no arguments" >&2
+    exit 1
+fi
+
+cd "$(dirname "$0")"/..
+
+find . -name Cargo.toml -exec cargo update --manifest-path {} \;
```
danielhenrymantilla, line 13: Note that there is a difference between a `Cargo.lock` being in "sync" with a `Cargo.toml` file (_i.e._, the versions picked in the `Cargo.lock` file are compatible with the constraints defined by the `Cargo.toml` file(s)), and a `Cargo.lock` being "up-to-date" with the latest crates.io release (_i.e._, the versions featured in the `Cargo.lock` are the _maximal_ compatible ones with the `Cargo.toml` constraints, _at that point in time_).

  - `cargo update -w` for the former,

  - `cargo update` for the latter.

While an up-to-date `Cargo.lock` is necessarily one in sync, I think it will make satisfying CI too demanding, since as time passes, new semver compatible versions of the dependencies may be released; and to properly acknowledge those releases, it's the `Cargo.toml` file which ought to be then updated.

That's why I may suggest using `cargo update -w` here instead, and maybe have a dependabot to handle the task of remaining up to date with crates.io releases 🙂 

```suggestion
find . -name Cargo.toml -exec cargo update -w --manifest-path {} \;
```

smoelius: The problem I most want to avoid is: needing a `Cargo.lock` to build and not knowing that you need it.

Do you have any thoughts on how best to achieve this?

danielhenrymantilla: The way I see it, the `Cargo.lock` file oughtn't be needed most of the time (say, 99% of the time #wavehanded), since that's what the whole semver thing is all about: with perfect semver and proper `Cargo.toml` files, `Cargo.lock` files would never be needed. So my whole suggestion (#30) is not about them being _mandatory_ to install `dylint`, to clarify.

They actually exist, imho, to offer a way to manage issues regarding semver breaking changes on some dependency, or slightly incorrect because under-constrained `Cargo.toml` files (_e.g._, using `syn = "1.0"` as incorrectly [preconised by `syn` itself](https://docs.rs/syn/1.0.73/syn/index.html#example-of-a-derive-macro), and then using [some API that isn't available on `1.0.0`](https://github.com/dtolnay/syn/releases/tag/1.0.1)). So there is that 1% of cases where people will be grateful you still had a `Cargo.lock`:

  - Breaking change from a dependency;

      - I remember having wanted to install `cargo-edit` at some point, and it had a dependency which had been incorrectly using `#[doc(hidden)] pub` unstable API from `serde`, and so when serde changed it, it was impossible to `cargo install` `cargo-edit`! But then many people said: "hey, if you add the `--locked` flag, you can still install it". Granted, it's not super common knowledge, but having the option to _make_ it compile until the maintainer amends its `Cargo.toml` files to circumvent the issue (if they ever do it), is a huge boon, and for the rare times where it is needed, knowledge of "the `--locked` trick" spreads.

  - Security-zealous people who want to guard against a supply-chain attack or some other sneaky patch.

Both could "work" with the `-Z minimal-version` flag of `cargo`, but then they'd stumble upon the common problem of people under-constraining their `Cargo.toml` files _w.r.t._ the minimal version required. Notice how, for instance, the very `syn` crate [shows a canonical example of `Cargo.toml` file with `syn = "1.0"`](https://docs.rs/syn/1.0.73/syn/index.html#example-of-a-derive-macro), even though `syn` has been continuously making API additions all along (even as of [`1.0.1`](https://github.com/dtolnay/syn/releases/tag/1.0.1)), which means an actual `=1.0.0` version pick is likely to cause breakage in many crates 😞 

___

Back to "the real world", what I personally do is have a "future-proofing" CI (using cron), which runs the tests for a _matrix_ combination of `--locked` / `cargo update`-d-beforehand and `MSRV/stable/beta`. The idea is that a "well maintained" crate ought to compile on all these environments (except, technically for `beta`, but the Rust maintainers would welcome the feedback) even with a maximal `Cargo.lock` (_i.e._, what a downstream would use if they don't use the `--locked` flag). That way one can ensure the `Cargo.lock` file oughtn't be needed.

But when doing `cargo bisect`s, or if the repo stops being maintained, I think having those `Cargo.lock` "snapshots" of a working dependency tree just _available_ can be immensely useful.

Sorry for the long messages, and I hope I have made sense 😄

smoelius: > Sorry for the long messages, and I hope I have made sense 😄

No worries at all. You have a lot of good ideas/suggestions, and I really appreciate the feedback. 🙏

PR #43, opened 2021-07-28
Allow cargo-dylint to be succesfully run on Windows.

MinerSebas:
This PR fixes `cargo-dylint` to run successfully on Windows.

Changes:
 - Usage of [`dirs::home_dir`](https://docs.rs/dirs/3.0.2/dirs/fn.home_dir.html) from the [dirs](https://crates.io/crates/dirs) crate to get the HOME directory.
 - Account for the fact that on windows the dylint driver is created with a file extension.
 - To successfully determine the driver Version, copies of the toolchain dll's were placed next to the driver.

`dylint-link` does not yet work windows. The correct library names for the examples are:
```
allow_clippy@nightly-2021-02-11-x86_64-pc-windows-msvc.dll
clippy@nightly-2021-04-08-x86_64-pc-windows-msvc.dll
env_literal@nightly-2021-04-08-x86_64-pc-windows-msvc.dll
path_separator_in_string_literal@nightly-2021-04-08-x86_64-pc-windows-msvc.dll
try_io_result@nightly-2021-04-22-x86_64-pc-windows-msvc.dll
```


smoelius:
Hey, @MinerSebas. This looks really good!

It looks like you're still pushing changes, though? Could you please ping me once you're happy with it?

MinerSebas:
@smoelius I am finished with my changes.
I had just a small churn due to the `unused-imports` lint, as the `read_dir` function was not necessary on Linux.

smoelius:
OK. Thanks. I'll start reviewing this as soon as possible.

dylint/src/driver_builder.rs
```diff
@@ -87,7 +90,7 @@ fn dylint_drivers() -> Result<PathBuf> {
         ensure!(dylint_drivers.is_dir());
         Ok(dylint_drivers.to_path_buf())
     } else {
-        let home = var(env::HOME)?;
+        let home = dirs::home_dir().ok_or_else(|| anyhow!("Could not find the HOME Directory."))?;
```
smoelius, line 93:
```suggestion
        let home = dirs::home_dir().ok_or_else(|| anyhow!("Could not find HOME directory"))?;
```
Nit to be consistent with error messages elsewhere.

dylint/src/driver_builder.rs
```diff
@@ -168,12 +171,37 @@ fn build(opts: &crate::Dylint, toolchain: &str, driver: &Path) -> Result<()> {
...
-            .join(format!("dylint_driver-{}", toolchain)),
+        package.join("target").join("debug").join(format!(
+            "dylint_driver-{}{}",
+            toolchain,
+            consts::EXE_SUFFIX
+        )),
         driver,
     )?;
 
+    // To succesfully determine the dylint driver Version on Windows,
```
smoelius, line 182:
```suggestion
    // MinerSebas: To succesfully determine the dylint driver Version on Windows,
```
I find it helpful to know who left comments. (I have yet to add a CONTRIBUTING.md file.)

internal/src/env.rs
```diff
@@ -9,7 +9,7 @@ pub const DYLINT_LIBRARY_PATH: &str = "DYLINT_LIBRARY_PATH";
 pub const DYLINT_LIBS: &str = "DYLINT_LIBS";
 pub const DYLINT_LIST: &str = "DYLINT_LIST";
 pub const DYLINT_RUSTFLAGS: &str = "DYLINT_RUSTFLAGS";
-pub const HOME: &str = "HOME";
+pub const HOME: &str = "HOME"; // Todo: Remove this on the next major semver change.
```
smoelius, line 12:
```suggestion
```
This package is meant to be internal to Dylint. So I have no problem removing this line now.

dylint/src/driver_builder.rs
```diff
@@ -168,12 +171,37 @@ fn build(opts: &crate::Dylint, toolchain: &str, driver: &Path) -> Result<()> {
...
+    // it is neccesary to place copies of the toolchain dll's next to the driver.
+    #[cfg(target_os = "windows")]
+    {
+        let path = PathBuf::new()
+            .join(env::var(env::RUSTUP_HOME)?)
```
smoelius, line 187:
```suggestion
        let rustup_home = var(env::RUSTUP_HOME)?;
        let path = PathBuf::from(rustup_home)
```
When an environment variable cannot be found,`dylint_internal::env::var` tells you which one.

dylint/src/driver_builder.rs
```diff
@@ -168,12 +171,37 @@ fn build(opts: &crate::Dylint, toolchain: &str, driver: &Path) -> Result<()> {
...
+            .join("toolchains")
+            .join(toolchain)
+            .join("bin");
+
+        for file in read_dir(path)?.flatten() {
+            let file_name = file.file_name();
+
+            if let Some(file_name) = file_name.to_str() {
+                if file_name.ends_with(".dll") {
```
smoelius, line 196:
```suggestion
                if file_name.ends_with(consts:: EXE_SUFFIX) {
```
I think a `use` declaration will need to be added.

smoelius (changes requested):
All my suggestions are minor. I really appreciate you doing this!

I have requested the help of someone with more Windows experience than me (@woodruffw) to also take a look.

woodruffw:
Took a brief look, no major objections to the code itself!

One thing to consider: it's possible to load the correct DLLs without having to copy them, by doing something tricky like this:

1. `cargo-dylint` is run by the user
2. `cargo-dylint` detects that it's on Windows
3. `cargo-dylint` re-spawns itself *in the toolchain directory*, with a special flag or environment variable telling it to change back to the package directory ASAP

This is consistent with Windows' convoluted DLL-load rules (since the re-spawned `cargo-dylint` would have have the correct `$CWD` for relative DLL discovery) and avoids a bunch of file copies. OTOH, it's a little bit harder to follow in terms of control flow and the driver already does a bunch of copying, so maybe it's not worth it. Just thought I'd put it out there!

smoelius:
Thanks a lot, @woodruffw.

Re re-spawning in the toolchain directory: I too think I am okay with the copying for now. We'll want to keep this idea in mind though. Thanks a lot for the suggestion.

@MinerSebas, if you could just review my suggestions, I think we should be good to go.

smoelius:
Thanks again, @MinerSebas.

PR #48, opened 2021-07-31
UI Testing fixes

MinerSebas:
When providing several Paths in DYLINT_LIBRARY_PATH, they were seperated by a `:`.
This broke UI Tests on Windows, as `C:\Dylint` would be split into `C` and `\Dylint`, neither of which are valid Windows Paths.

When testing examples as a UI Test, the deletion of the previous Veriosn didn't check the file extension and was thus broken on windows.

### DIFF
--- a/cargo-dylint/tests/list.rs
+++ b/cargo-dylint/tests/list.rs
@@ -3,7 +3,7 @@ use assert_cmd::prelude::*;
 use dylint_internal::{cargo::SanitizeEnvironment, env, Command};
 use predicates::prelude::*;
 use std::{
-    ffi::OsString,
+    env::join_paths,
     path::{Path, PathBuf},
 };
 use tempfile::tempdir;
@@ -90,11 +90,11 @@ fn one_name_multiple_paths() {
         .success()
         .unwrap();
 
-    // smoelius: https://users.rust-lang.org/t/osstring-osstr-error/35249
-    let mut paths = OsString::new();
-    paths.push(&target_debug(tempdirs.0.path()));
-    paths.push(";");
-    paths.push(&target_debug(tempdirs.1.path()));
+    let paths = join_paths(&[
+        &target_debug(tempdirs.0.path()),
+        &target_debug(tempdirs.1.path()),
+    ])
+    .unwrap();
 
     std::process::Command::cargo_bin("cargo-dylint")
         .unwrap()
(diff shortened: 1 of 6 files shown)


MinerSebas:
Found another Bug in the UI Testing.
When trying it inside CI, the Test always failed because it set CARGO_TERM_COLOR to "always".
The Regex that parses the Command Output didn't account for this, which prevented the discovery if the rustic flags.
Now it will always set that env var too "never",

smoelius:
> Found another Bug in the UI Testing.
> When trying it inside CI, the Test always failed because it set CARGO_TERM_COLOR to "always".

I'm confused because this seems to work in the current CI. What would cause this bug to trigger?

Separately, are you still pushing to this branch?

MinerSebas:
> > Found another Bug in the UI Testing.
> > When trying it inside CI, the Test always failed because it set CARGO_TERM_COLOR to "always".
> 
> I'm confused because this seems to work in the current CI. What would cause this bug to trigger?

Here is a Failing CI, with extra `dbg!()` thrown in that shows the Command Outoput with the Color Characters in the Lines.
Your CI doesn't set CARGO_TERM_COLOR , which means it defaults to "auto" which doesn't add the Color Characters:
https://doc.rust-lang.org/cargo/reference/config.html#termcolor

> Separately, are you still pushing to this branch?

This should be all now that bevyengine/bevy #2582 doesn't fail its CI.

MinerSebas:
Sorry forgot to actually Link the CI run: https://github.com/MinerSebas/bevy/runs/3215084903?check_suite_focus=true#step:10:140

smoelius:
> Your CI doesn't set CARGO_TERM_COLOR , which means it defaults to "auto" which doesn't add the Color Characters:
> https://doc.rust-lang.org/cargo/reference/config.html#termcolor

Ah. I understand now. Thanks.

> This should be all, now that bevyengine/bevy#2582 doesn't fail its CI.

OK. Thanks. I think I will have some small changes for this PR as well. I will probably push to your branch and then ask you to review them.

smoelius:
It turns out there is a portable way to handle path delimiters: `join_paths` and `split_paths`.

Using those was the main adjustment that I made.

When you have a moment, @MinerSebas, please take a look at 612369e22d201701d2b474875dad43250d831763 and let me know if it looks okay to you.

MinerSebas:
@smoelius The changes work without any Problems. 👍

smoelius:
Cool. Thanks a lot, @MinerSebas. I'll get to the next PR soon.

Issue #36, opened 2021-06-23
Set up recurring task to verify `Cargo.lock` dispensability


smoelius:
In #34, I asked:
> The problem I most want to avoid is: needing a `Cargo.lock` to build and not knowing that you need it.
> 
> Do you have any thoughts on how best to achieve this?

And @danielhenrymantilla respondend:
> Back to "the real world", what I personally do is have a "future-proofing" CI (using cron), which runs the tests for a _matrix_ combination of `--locked` / `cargo update`-d-beforehand and `MSRV/stable/beta`. The idea is that a "well maintained" crate ought to compile on all these environments (except, technically for `beta`, but the Rust maintainers would welcome the feedback) even with a maximal `Cargo.lock` (_i.e._, what a downstream would use if they don't use the `--locked` flag). That way one can ensure the `Cargo.lock` file oughtn't be needed.

smoelius:
After [modifying](https://github.com/trailofbits/dylint/commit/f71c4d917f5ae0433cb1c83ca930178f41c8f305) #50 to use a deploy key, it seems to be working.

Issue #35, opened 2021-06-23
Investigate Dependabot


smoelius:
In #34, @danielhenrymantilla wrote:
> That's why I may suggest using `cargo update -w` here instead, and maybe have a dependabot to handle the task of remaining up to date with crates.io releases 🙂

smoelius:
https://github.com/trailofbits/dylint/commit/e5d113ad2ea1431c82a52c5b10b34257d16dab1c

smoelius:
Closing as [this](https://github.com/trailofbits/dylint/blob/master/.github/dependabot.yml) seems to be working.

PR #50, opened 2021-07-31
GitHub action to update dependencies

smoelius:
Addresses #36. Obviates #35 (I think).


smoelius:
> Obviates #35 (I think).

This is not correct. This action will update patch versions, but not major or minor versions.

PR #57, opened 2021-08-05
Add Toolchain Folder to Path, instead of copying dll's.

MinerSebas:
In #43 one of the fixes was to copy Libraries to the Driver Folder.
This PR instead provides those Libraries by adding them to the Path that's supplied when determining the Driver Version.

This doesn't have to be merged, its just an alternate Solution I wanted to bring up.


smoelius:
Thanks, @MinerSebas. I like this approach much better. I'll get to this shortly.

MinerSebas:
I was dissatisfied with how I had to disable Lints for non-windows platforms.
For this reason, I rewrote the change using `cfg!` instead.

This should now be the final version.

PR #58, opened 2021-08-05
In the VS Code Integration also include the --all-targets flag

MinerSebas:
This runs dylint also on Benches, Examples and Tests.


smoelius:
Thanks!

PR #60, opened 2021-08-06
Have all workspaces share a target directory

smoelius:
This has numerous benefits:

* Build times are faster, because all of the workspaces share a `deps` directory.
* `cargo clean` now works.
* It simplifies the script used to lint the repository.

Thanks to @thomaseizinger for showing us this was possible.


thomaseizinger:
Oh that is exciting!

Thank you!

PR #65, opened 2021-08-16
Fix Clap Dependency

MinerSebas:
The latest update of Clap breaks the installation of cargo-dylint ([Concrete Case](https://github.com/trailofbits/dylint-template/pull/6/checks?check_run_id=3333730574)).

The problem was fixed by removing the `multiple = true,` inside the derive annotations.
I also pinned the clap dependency to avoid further breakage. (Cargo sees Beta releases as Semver compatible)

A downside to the move to Clap `3.0.0-beta.4` is that it uses the [Attributes can invoke function-like macros](https://blog.rust-lang.org/2021/07/29/Rust-1.54.0.html#attributes-can-invoke-function-like-macros) Feature of Rust 1.54, which prevents the `allow-clippy` Lint from linting cargo-dylint correctly.

Either `allow-clippy` should be moved to a more recent rust version or pin clap to `3.0.0-beta.2` instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,9 +37,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.42"
+version = "1.0.43"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "595d3cfa7a60d4555cb5067b99f07142a08ea778de5cf993f7b75c7d8fabc486"
+checksum = "28ae2b3dec75a406790005a200b1bd89785afc02517a00ca99ecfe093ee9e6cf"
 
 [[package]]
 name = "assert_cmd"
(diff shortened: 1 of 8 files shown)


smoelius1:
Hi, @MinerSebas. Thanks a lot for looking into this!

I am currently travelling (hence the funny account).

I will take a look at this as soon as I get back (in about a week).

smoelius:
@MinerSebas, I really, genuinely appreciate you digging into this.

I don't think we want to remove `multiple` entirely, as that would cause invocations like the following to fail:
```
cargo-dylint --lib allow_clippy --lib env_literal
```
I found this in Clap's changelog (https://github.com/clap-rs/clap/blame/master/CHANGELOG.md#L31-L34):

* **Removed Methods**
  * **Arg**
    * `Arg::settings` in favor of `Arg::setting(Setting1 | Setting2)`
    * `Arg::multiple` in favour of `Arg::multiple_values` and `Arg::multiple_occurrences`

Based on my experiments, it looks like `multiple_occurrences` is the one we want.

Also, re pinning Clap, I understand why you've suggested this, but I think I would prefer not to, because I like to know when changes to dependencies cause things to break. I'm not 100% sure this is the right approach, but it's what I'm going with for now.

Please let me know if you have strong objections to anything I've said.

And thanks again for doing all the legwork on this.

smoelius:
Actually, I am going to go ahead and merge this so that the lockfiles bot doesn't fail again. (It failed last week because of the clap issue.)

Please don't feel like you can't still comment, because you are more than welcome to.

PR #78, opened 2021-09-11
Hide dylint_version() in docs

MinerSebas:
While uploading the Docs for my Crate, I noticed that I can't hide the `dylint_version()` Function inside the docs.
https://minersebas.github.io/bevy_lint/bevy_lint/index.html#functions

This adds the `#[doc(hidden)]` Attribute inside the `dylint_library!` Macro, to hide the function.


smoelius:
I obsessive-compulsively sorted those attribute macros. But this definitely seems like a good idea.

Your help is always appreciated, @MinerSebas.

PR #170, opened 2022-01-01
Add experimental `--bisect` option

smoelius:


dylint/src/package_options/bisect.rs
```diff
@@ -0,0 +1,132 @@
...
+if [[ ! -f first_channel_seen ]]; then
+    cargo build --tests || (touch first_channel_seen && false)
+elif [[ ! -f successful_build_seen ]]; then
+    # smoelius: Pretend the build succeeds (even if it doesn't) and proceed backwards until we find
+    # one that actually succeeds.
+    (cargo build --tests && touch successful_build_seen) || true
+else
+    cargo build --tests
+fi
+"#;
```
jyn514, line 53: An easy way to make this cross-platform like this is something like https://github.com/matklad/xshell/, which has mostly the same syntax as shell, but uses `Command::new` under the hood and doesn't depend on a specific shell.

jyn514: Oh, I just saw you're passing the path to the script as an argument to some tool - maybe you could do that with xshell by adding a second binary and passing its path? something like `src/bin/bisect.rs` and then `cargo install` would install it alongside cargo-dylint.

smoelius: I think I like this idea.

However, given that the script mostly works and I have other priorities ([cough](https://github.com/rust-lang/rust/issues/66079#issuecomment-1005749644)), I will likely merge this as-is and make implementing your suggestion a TODO item.

Thanks, BTW.

Issue #337, opened 2022-05-30
Incompatibility with mold linker


brson:
I recently tried to `cargo install` a crate (`cargo-contract`) that relies on `dylint-link` and it failed due to interactions with the recommended configuration for using the [mold linker](https://github.com/rui314/mold).

Mold is configured by putting this in `~/.cargo/config.toml`:

```toml
[target.x86_64-unknown-linux-gnu]
linker = "clang"
rustflags = ["-C", "link-arg=-fuse-ld=/path/to/mold"]
```

This is a global configuration saying to use `clang` as the linker, and clang understands the `-fuse-ld` argument.

`dylint-link` though invokes `cc`, which on my system is `gcc` and I see the following error:

```
     Compiling memchr v2.4.1
  error: linking with `dylint-link` failed: exit status: 1
  ...
    = note: cc: error: unrecognized command line option '-fuse-ld=/usr/local/bin/mold'
```

It's not immediately obvious to me how to change `dylint-link` to handle this situation.

smoelius:
I think `dylint-link` could check for the existence of `~/.cargo/config.toml`, and whether there is an applicable setting in there.

I.e., if something like this appears, then invoke `clang` instead of `cc`:
```toml
[target.x86_64-unknown-linux-gnu]
linker = "clang"
```

Does that sound reasonable?

brson:
That sounds like it would work for my use case at least, and a good start. It may not be foolproof - there may for example be other locations where cargo might be looking for `.cargo/config`, or otherways to communicate to rustc/cargo which linker to use.

Thanks for the quick response.

brson:
@smoelius yes, the new version works for me. Here is what I ran

```
cargo install dylint-link cargo-dylint
cargo install cargo-contract
```

Thanks!

smoelius:
Thank you!

PR #339, opened 2022-06-01
Respect `$CARGO_HOME/config.toml` in `dylint-link`

smoelius:


smoelius:
I think this was the fix for #337.

PR #477, opened 2022-10-29
Allow registering tool lints

bb010g:
Existing lints work as before. Example use:
```rust
dylint_linting::declare_late_lint! {
    /// …
    pub dylint::QUESTION_MARK_IN_EXPRESSION,
    Warn,
    "`?` operators embedded within an expression"
}
```

### DIFF
--- a/utils/linting/src/lib.rs
+++ b/utils/linting/src/lib.rs
@@ -23,6 +23,20 @@ macro_rules! dylint_library {
 
 #[macro_export]
 macro_rules! __declare_and_register_lint {
+    ($(#[$attr:meta])* $vis:vis $tool:ident :: $NAME:ident, $Level:ident, $desc:expr, $register_pass_method:ident, $pass:expr) => {
+        $crate::dylint_library!();
+
+        extern crate rustc_lint;
+        extern crate rustc_session;
+
+        #[no_mangle]
+        pub fn register_lints(_sess: &rustc_session::Session, lint_store: &mut rustc_lint::LintStore) {
+            lint_store.register_lints(&[$NAME]);
+            lint_store.$register_pass_method($pass);
+        }
+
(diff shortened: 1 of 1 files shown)


smoelius:
Thanks very much for your PR.

I am okay with adding this functionality, and I like the idea of having one set of macros that handle both tool and non-tool lints, but I don't know how I feel about this:
https://github.com/trailofbits/dylint/blob/6853c35630431c90984565c165b43de3f93cf820/utils/linting/src/lib.rs#L175
Specifically, introducing special syntax (`[@tool]`) into a macro a user is expected to use.

What I am about to propose is not ideal, but I think I would prefer have two (nearly identical) alternatives within the macro definition, similar to what you did for [`_declare_and_register_lint`](https://github.com/trailofbits/dylint/blob/6853c35630431c90984565c165b43de3f93cf820/utils/linting/src/lib.rs#L25).

Also, we will definitely want a test for this new functionality, e.g., something like:
- create a new Dylint template in a temporary directory
- use [`dylint_internal::find_and_replace`](https://github.com/trailofbits/dylint/blob/c458ffa63df682b66154504061bc7b99da2cf198/internal/src/sed.rs#L7) to add a tool name 
- verify that the modified code builds and the UI tests pass

The [`boundary_toolchains`](https://github.com/trailofbits/dylint/blob/master/cargo-dylint/tests/boundary_toolchains.rs) test could probably be used as a starting point.

smoelius (changes requested):
https://github.com/trailofbits/dylint/pull/477#issuecomment-1296851216

smoelius:
@bb010g Do you think you might continue to work on this?

smoelius:
@bb010g I'm going to close this for bookkeeping purposes. But I would be happy to revisit this, especially if you have a need for it.

PR #720, opened 2023-06-13
Use ctor to set the current directory in the dylint.rs tests

### LINKED ISSUE #718: Use `ctor` to set the current directory in the dylint.rs tests
To reduce the proliferation of `..` in that file (e..g, [here](https://github.com/trailofbits/dylint/blob/7ab1516015549dbadc277cc2fe2fcb0d71170388/cargo-dylint/tests/dylint.rs#L36), [here](https://github.com/trailofbits/dylint/blob/7ab1516015549dbadc277cc2fe2fcb0d71170388/cargo-dylint/tests/dylint.rs#L70), [here](https://github.com/trailofbits/dylint/blob/7ab1516015549dbadc277cc2fe2fcb0d71170388/cargo-dylint/tests/dylint.rs#L305), and [here](https://github.com/trailofbits/dylint/blob/7ab1516015549dbadc277cc2fe2fcb0d71170388/cargo-dylint/tests/dylint.rs#L323)).

And example where `ctor` is used in this way: https://github.com/trailofbits/necessist/blob/7d46abec83529e449a6b7fbc7bacf73b02e1079c/core/tests/ci.rs#L17

nachiketkanore:
- fixes #718


smoelius:
Hi, @nachiketkanore. Thank you very much for the PR!

Dylint's CI is pretty nit-picky. Please let me know if you have any questions on how to get the tests to pass.

nachiketkanore:
Thanks. I probably need to better understand how the tests are being run.
As I can see from the logs, some directory names were unidentified!
I will go through this thoroughly..

nachiketkanore:
I am unable to determine the cause of failing tests in CI
Locally, the tests are passing 
(I am running this command `$ cargo test -p cargo-dylint --test dylint`)

smoelius:
> Also, locally it only runs 11 tests which are all passing but in CI, there is one extra test named `hack_feature_powerset` which I am unable to navigate anywhere. Any lead is appreciated. Thanks

I threw you a curve ball. That test was added here: https://github.com/trailofbits/dylint/pull/717

Could please you try rebasing?

cargo-dylint/tests/dylint.rs
```diff
@@ -329,7 +335,7 @@ fn compare_lines(left: &str, right: &str) {
 // smoelius: Skip examples directory for now.
 fn walkdir(include_examples: bool) -> impl Iterator<Item = walkdir::Result<walkdir::DirEntry>> {
     #[allow(unknown_lints, env_cargo_path)]
-    walkdir::WalkDir::new(Path::new(env!("CARGO_MANIFEST_DIR")).join(".."))
+    walkdir::WalkDir::new(Path::new(current_dir().unwrap().to_str().unwrap()))
```
smoelius, line 338:
```suggestion
    walkdir::WalkDir::new(".")
```
`WalkDir::new`'s argument must implement [`AsRef<Path>`](https://docs.rs/walkdir/latest/walkdir/struct.WalkDir.html#method.new), which `&str` [does](https://doc.rust-lang.org/std/primitive.str.html#impl-AsRef%3CPath%3E-for-str), so I think this ought to work.

cargo-dylint/tests/dylint.rs
```diff
@@ -7,7 +7,10 @@ use regex::Regex;
 use sedregex::find_and_replace;
 use semver::Version;
 use similar_asserts::SimpleDiff;
+use std::env::current_dir;
```
smoelius, line 10:
```suggestion
```
With the changes below, I think you ought to be able to get rid of this.

cargo-dylint/tests/dylint.rs
```diff
@@ -7,7 +7,10 @@ use regex::Regex;
...
 use std::{
+    env,
```
smoelius, line 12:
```suggestion
```
Not sure why this was needed.

smoelius (changes requested):
I think we're nearing the end! Thanks for all of you hard work on this!

cargo-dylint/tests/dylint.rs
nachiketkanore: This fails the `markdown_link_check` test, as it doesn't create full paths
`Error: ENOENT: no such file or directory, stat \'./cargo-dylint/README.md\'

smoelius: Ah. That is arguably a flaw in the `markdown_link_check` test. But I can fix that in a subsequent PR.

smoelius (approved):
Thanks very much for this!

nachiketkanore:
Thanks for helping me out in this! I am still a beginner, sorry for any inconvenience caused!

smoelius:
No inconvenience at all. I really appreciate your help!

PR #750, opened 2023-07-13
Remove unnecessary `RefCell` usage

EFanZh:
It may be not necessary to use `RefCell` here.

A better way could be using [`OnceCell`](https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html) or [`OnceLock`](https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html), since we can get a reference instead of cloning toml values, which could be a little faster.


smoelius:
> It may be not necessary to use `RefCell` here.
> 
> A better way could be using [`OnceCell`](https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html) or [`OnceLock`](https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html), since we can get an reference instead of cloning toml values, which could be a little faster.

Just to be sure I understand:
- This PR remove the unnecessary use of `RefCell`.
- On top of that, we might do away with the `Mutex` entirely, and use a `OnceCell` or `OnceLock` instead.

Is that right?

Aside: the CI failure is because your PR happens to coincide with the Rust 1.71.0 release. I opened #751 to fix that. I'll ping you once that's merged, if you wouldn't mind rebasing.

smoelius:
Please rebase when you have a moment. I expect CI to pass for your PR now.

EFanZh:
> > It may be not necessary to use `RefCell` here.
> > A better way could be using [`OnceCell`](https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html) or [`OnceLock`](https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html), since we can get an reference instead of cloning toml values, which could be a little faster.
> 
> Just to be sure I understand:
> 
> * This PR removes the unnecessary use of `RefCell`.
> * On top of that, we might do away with the `Mutex` entirely, and use a `OnceCell` or `OnceLock` instead.
> 
> Is that right?
> 
> Aside: the CI failure is because your PR happens to coincide with the Rust 1.71.0 release. I opened #751 to fix that. I'll ping you once that's merged, if you wouldn't mind rebasing.

Yes, that is what I mean.

smoelius:
Thanks very much! This is a definite improvement!

PR #997, opened 2024-01-05
Disable `custom_toolchain` test when running under `cargo-llvm-cov`

smoelius:


cargo-dylint/tests/custom_toolchain.rs
```diff
@@ -1,6 +1,8 @@
 // smoelius: As per `dylint-link/src/main.rs`:
 // "Only the MSVC toolchain is supported on Windows"
 #![cfg(not(target_os = "windows"))]
+// smoelius: I haven't been able to figure out why this test fails when run under `cargo-llvm-cov``.
```
smoelius, line 4:
```suggestion
// smoelius: I haven't been able to figure out why this test fails when run under `cargo-llvm-cov`.
```

PR #1015, opened 2024-01-12
Experiment with collapsible logs

smoelius:


smoelius:
This experiment has taught me the following. Only "boring" things should be collapsed, e.g., build logs. Things like lint warnings should be visible without having to click.

PR #1019, opened 2024-01-17
Add function name to the non local effect lint

fcasal:
This PR adds the called function formatted string to the non-local-effect lint message. 
This allows knowing which function triggered the lint if the issue is in a derived/expanded function (e.g., in a structure declaration).

### DIFF
--- a/examples/general/non_local_effect_before_error_return/src/lib.rs
+++ b/examples/general/non_local_effect_before_error_return/src/lib.rs
@@ -224,7 +224,7 @@ fn is_call_with_mut_ref<'tcx>(
         if locals.iter().any(|local| is_mut_ref_arg(mir, local))
             || constants.iter().any(|constant| is_const_ref(constant));
         then {
-            let func_string = format!("{:#?}", func);
+            let func_string = format!("{func:#?}");
             Some(FunctionStringAndSpan(func_string, *fn_span))
         } else {
             None
--- a/examples/general/non_local_effect_before_error_return/src/lib.rs
+++ b/examples/general/non_local_effect_before_error_return/src/lib.rs
@@ -150,22 +150,21 @@ impl<'tcx> LateLintPass<'tcx> for NonLocalEffectBeforeErrorReturn {
             |path, contributing_calls, span| {
                 // smoelius: The path is from a return to the start block.
                 for (i, &index) in path.iter().enumerate() {
-                    let basic_block = &mir.basic_blocks[index];
-
                     if_chain! {
                         if !contributing_calls.contains(index);
-                        if let Some(call_span) = is_call_with_mut_ref(cx, mir, &path[i..]);
+                        if let Some(func_and_span) = is_call_with_mut_ref(cx, mir, &path[i..]);
                         then {
                             span_lint_and_then(
                                 cx,
                                 NON_LOCAL_EFFECT_BEFORE_ERROR_RETURN,
-                                call_span,
-                                "call with mutable reference before error return",
+                                func_and_span.1,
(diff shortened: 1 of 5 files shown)


examples/general/non_local_effect_before_error_return/src/lib.rs
```diff
@@ -150,22 +150,21 @@ impl<'tcx> LateLintPass<'tcx> for NonLocalEffectBeforeErrorReturn {
             |path, contributing_calls, span| {
                 // smoelius: The path is from a return to the start block.
                 for (i, &index) in path.iter().enumerate() {
-                    let basic_block = &mir.basic_blocks[index];
-
                     if_chain! {
                         if !contributing_calls.contains(index);
-                        if let Some(call_span) = is_call_with_mut_ref(cx, mir, &path[i..]);
+                        if let Some(func_and_span) = is_call_with_mut_ref(cx, mir, &path[i..]);
```
smoelius, line 155:
```suggestion
                        if let Some((func, span)) = is_call_with_mut_ref(cx, mir, &path[i..]);
```
Style nit (saves having to write `.0` and `.1`). Note that the code in the block would need to be adjusted.

examples/general/non_local_effect_before_error_return/src/lib.rs
```diff
@@ -150,22 +150,21 @@ impl<'tcx> LateLintPass<'tcx> for NonLocalEffectBeforeErrorReturn {
...
                         then {
                             span_lint_and_then(
                                 cx,
                                 NON_LOCAL_EFFECT_BEFORE_ERROR_RETURN,
-                                call_span,
-                                "call with mutable reference before error return",
+                                func_and_span.1,
+                                format!("call to {} with mutable reference before error return", func_and_span.0).as_str(),
```
smoelius, line 161: Style nit: `&format!(...)` is more concise than `format!(...).as_str()`.

examples/general/non_local_effect_before_error_return/src/lib.rs
```diff
@@ -223,7 +224,8 @@ fn is_call_with_mut_ref<'tcx>(
         if locals.iter().any(|local| is_mut_ref_arg(mir, local))
             || constants.iter().any(|constant| is_const_ref(constant));
         then {
-            Some(*fn_span)
+            let func_string = format!("{func:#?}");
```
smoelius, line 227: It feels odd to me to convert something to a string using `Debug` formatting, only to later print it with `Display` formatting.

Could we instead change the return type to `Option<(&'a Operand, Span)>`, and include the operand directly in the warning using `Debug` formatting?

Also, I'm not certain, but I think just `{:?}` might work (i.e., instead of `{:#?}`).

examples/general/non_local_effect_before_error_return/ui/main.stderr
```diff
@@ -12,7 +12,7 @@ LL |     Err(VarError::NotPresent)
    = note: `-D non-local-effect-before-error-return` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(non_local_effect_before_error_return)]`
 
-error: call with mutable reference before error return
+error: call to std::vec::Vec::<u32>::push with mutable reference before error return
```
smoelius, line 15:
```suggestion
error: call to `std::vec::Vec::<u32>::push` with mutable reference before error return
```
Since the function paths are technically syntax, could we enclose them in backticks (as shown)?

smoelius (commented):
Thanks a lot for doing this, @fcasal. My comments are all very minor and nit-picky.

Do you think it would be difficult to include an example in the tests of why this was desired in the first place? You can say no, we can merge this, and leave that for another day.

PR #1032, opened 2024-01-22
Adds the basic-dead-store lint

fcasal:


examples/general/basic_dead_store/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "basic_dead_store"
+version = "0.1.0"
```
smoelius, line 3:
```suggestion
version = "2.6.1"
```

examples/general/basic_dead_store/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+authors = ["Filipe Casal <fcasal@users.noreply.github.com>"]
+description = "A lint to find simple instances of dead stores in arrays."
+edition = "2021"
+publish = false
+
+[lib]
+crate-type = ["cdylib"]
```
smoelius, line 10:
```suggestion
crate-type = ["cdylib", "rlib"]
```
Both are needed, the latter for when all of the "general" lints are put into one `general` library.

examples/general/basic_dead_store/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "basic_dead_store"
+version = "0.1.0"
+authors = ["Filipe Casal <fcasal@users.noreply.github.com>"]
+description = "A lint to find simple instances of dead stores in arrays."
```
smoelius, line 5:
```suggestion
description = "A lint to find simple instances of dead stores in arrays"
```

examples/general/basic_dead_store/README.md
```diff
@@ -0,0 +1,24 @@
+# literal-indexed-dead-store
```
smoelius, line 1:
```suggestion
# basic_dead_store
```

examples/general/basic_dead_store/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/// Checks if the given expression is an assignment to an array indexed by a literal.
+/// Returns the tuple (array name, indexed position, span)
+fn is_assignment_to_array_indexed_by_literal(
+    expr: &Expr,
+    arr_string: &String,
+    tcx: &LateContext<'_>,
+) -> Option<(String, u128, Span)> {
+    let index_expr = get_parent_expr(tcx, expr)?;
```
smoelius, line 77: If I'm understanding correctly, this function matches an index expression, then it matches _upward_ to see if it's in an assignment, then it matches _downward_ to see if the index is a literal. Is that right?

Would it make more sense to match the assignment and then just match downward?

I'm not sure if this is helpful: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=98acb5ca4f29b9d2be7d69fece5a77ed

examples/general/basic_dead_store/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+    if let ExprKind::Index(array, index, _span) = index_expr.kind {
```
smoelius, line 78: I think this would be more clear written using an [if-let-chain](https://rust-lang.github.io/rfcs/2497-if-let-chains.html).

I think this will require adding:
```rust
#![feature(let_chains)]
```

examples/general/basic_dead_store/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+        if array.hir_id == expr.hir_id {
+            let assign_expr = get_parent_expr(tcx, index_expr)?;
+            if let ExprKind::Assign(target, _value, assignment_span) = assign_expr.kind {
+                if target.hir_id == index_expr.hir_id {
+                    if let ExprKind::Lit(lit) = index.kind {
+                        if let LitKind::Int(index, _type) = lit.node {
+                            return Some((arr_string.to_string(), index, assignment_span));
```
smoelius, line 85: We could leave the first component a `String`, but would it make more sense to return an `&Expr` and to compare them using `clippy_util`'s [`eq_expr`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/struct.SpanlessEq.html#method.eq_expr)?

This code is dated, but here is an example of `eq_expr` being used: https://github.com/crytic/solana-lints/blob/3c7044bc874ec7aeaee93e0fcf88a10793057bd4/lints/missing_owner_check/src/lib.rs#L111-L112

examples/general/basic_dead_store/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+                        "reassigning the same array position without using it",
+                        Some(in_common.first().unwrap().2),
+                        "original assignment was here",
+                    );
+                }
+            } else {
+                // If we are using the array in a way that is not an assignment to a certain
+                // position, then it is being used. Therefore, we need to clear all
+                // stored instances of this array
+                self.clear_stores_of(&arr_string);
```
smoelius, line 135: No action is needed, but this gives me pause, because it makes assumptions about the order in which expressions are processed.

I suspect those assumptions hold in practice, but it would be nice if we had a solution that didn't rely on them.

examples/general/basic_dead_store/src/lib.rs
```diff
@@ -0,0 +1,147 @@
+#![feature(rustc_private)]
+#![warn(unused_extern_crates)]
+
+extern crate rustc_ast;
+extern crate rustc_hir;
+extern crate rustc_span;
+use clippy_utils::{diagnostics::span_lint_and_help, get_parent_expr};
+
```
smoelius, line 8:
```suggestion

use clippy_utils::{diagnostics::span_lint_and_help, get_parent_expr};
```
Style nit (combine this `use` with the ones just below it).

smoelius (commented):
Your comments made this real easy to review. But please forgive me if I need to make a second pass.

examples/general/basic_dead_store/src/lib.rs
fcasal: I agree -- initially, I was checking on functions and iterating over the statements of the basic blocks but I was having issues finding all nested expressions within.

examples/general/basic_dead_store/src/lib.rs
fcasal: Yes, currently, we start with expr and need to check all other 3 nodes:
```
       assign_expr is ExprKind::Assign
         /
     index_expr is ExprKind::Index
   /         \
expr        lit is ExprKind::Lit
```
Matching on `expr` and going up and down in the tree seemed better:
Let's suppose I found all assignments to arrays with literals. Then, we would need to determine if there are any uses of these arrays between the assignments to the same position.
To determine if a Path expression of the same array was a `use`, we would need to check that it wasn't the original assignment and that it was in the proper location (i.e., between the previously collected assignments). This seemed much more complex than navigating at the path level.

Navigating at the Path level meant that I would never see the same array reference twice, and the check_expr return order seemed to match the program order.

smoelius: > To determine if a Path expression of the same array was a `use`, we would need to check that it wasn't the original assignment

OK. I see what you are saying.

examples/general/basic_dead_store/src/lib.rs
fcasal: I simplified the logic in the latest commit -- there was an unnecessary argument passing. Saving expr's required me to handle lifetimes on the BasicDeadStore struct and I felt that comparing the strings was fine. Do you think it is worth using eq_expr?

examples/general/basic_dead_store/src/lib.rs
fcasal: I agree -- and this is much better than if_chain!

examples/general/basic_dead_store/src/lib.rs
smoelius: Actually, there may be a better way still, using `expr_local`: https://doc.rust-lang.org/stable/nightly-rustc/clippy_utils/mir/fn.expr_local.html

But we can leave it as strings for now. That way we have a set of working tests to compare a new approach to.

PR #1159, opened 2024-05-03
Move `overscoped_allow` to experimental

smoelius:


smoelius:
I've decided that the churn this would cause for all of the workspaces that use this lint is not worth it.

PR #1508, opened 2025-02-04
Make the path argument to optional

### LINKED ISSUE #601: Make the path argument to `cargo dylint upgrade` optional
Make this optional:
```
$ cargo dylint upgrade --help
Upgrade the library package at <PATH> to the latest version of `clippy_utils`

Usage: cargo-dylint dylint upgrade [OPTIONS] <PATH>
                                             ^^^^^^
```
Default to the current directory when `PATH` is not provide.

The changes that would be required are the following:
- Change `path`'s type from `String` to `Option<String>` in two places:
  - https://github.com/trailofbits/dylint/blob/d75ae91f34613a329a231797a0142066194b5f3d/cargo-dylint/src/main.rs#L128
  - https://github.com/trailofbits/dylint/blob/d75ae91f34613a329a231797a0142066194b5f3d/dylint/src/opts.rs#L110
- Handle the case when `path` is `None` here: https://github.com/trailofbits/dylint/blob/d75ae91f34613a329a231797a0142066194b5f3d/dylint/src/package_options/mod.rs#L100

YanVictorSN:
## Description:

* [X] Changed the Change path's type from String to Option<String> in two places
* [X] Handle the case when path is None.

I also have to handle path in 
```rust
pub fn auto_correct_revertible()
````
function.

Close #601 

@smoelius

### DIFF
--- a/cargo-dylint/src/main.rs
+++ b/cargo-dylint/src/main.rs
@@ -125,7 +125,7 @@ Combine with `--all` to list all lints in all discovered libraries."
         rust_version: Option<String>,
 
         #[clap(help = "Path to library package")]
-        path: String,
+        path: Option<String>,
 
         #[clap(
             help_heading = Some("Experimental"),
--- a/dylint/src/opts.rs
+++ b/dylint/src/opts.rs
@@ -109,7 +109,7 @@ pub struct Upgrade {
 
     pub auto_correct: bool,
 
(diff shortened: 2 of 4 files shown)


smoelius:
Thanks, @YanVictorSN !

YanVictorSN:
@smoelius. Thank you!

PR #1517, opened 2025-02-09
Eliminate uses of `once_cell::sync::Lazy`

smoelius:


smoelius:
Hold off until Dylint's MSRV is 1.80 or later.

Issue #1225, opened 2024-05-24
Lint idea: nonexistent path in comment


smoelius:
Like this one, which I am fixing now: https://github.com/trailofbits/dylint/blob/f8cb8ed1e1d9cb8af615f1edf153061d7fcb9022/dylint/src/lib.rs#L22

smoelius:
Implemented by #1523

PR #1523, opened 2025-02-12
feat(examples)#1225: non existent path in comment

augustin-v:
### Description
Hello, this PR introduces a lint that checks comments in the code for file path references and verifies that the referenced files actually exist; emitting a warning when a comment points to a missing file.

### Problems encountered
Current implementation only supports checking parts that are absolute or relative to the file in which they are written. 
Also the file references from the project root might be incorrectly flagged even if they exist, e.g.:
```rust
// dylint/dylint/Cargo.toml
```
would get falsely flagged because the lint cannot currently determine which directory is the project root.


smoelius:
Thank you very much for this!

I'll give a more thorough review within the coming days, but I have two questions right now.

First, the general approach is to read each file in its entirety, then use regular expressions to find the comments, and then use regular expressions to find paths within the comments. Is that right?

Second, could you please describe in plain English what this regex matches? https://github.com/trailofbits/dylint/blob/b46bf0d7c9eac9193d268cbc901f8701a283a2a9/examples/supplementary/nonexistent_path_in_comment/src/lib.rs#L40

augustin-v:
Hi @smoelius thank you for the feedback!

> First, the general approach is to read each file in its entirety, then use regular expressions to find the comments, and then use regular expressions to find paths within the comments. Is that right?

Yes that is correct. It reads each file entirely, uses separate regular expressions to extract both line and block comments, and then applies another regular expression to find paths within those comments.


> Second, could you please describe in plain English what this regex matches?

The pattern `((?:\.\./|/|[\w/-]+/)+[\w-]+(?:\.[\w-]+)+)` is designed to match:
1. One or more directory segments, which can be:
 - "../" (parent directory) or "/" (root directory), 
 - A directory name ending with "/"
 2. followed by a file name (`[\w-]+` in the middle of the expression)
 3. The last part of the regular expression matches with one or more file extensions (each starting with a dot) in order to be able to support paths to files like `path/to/file.tar.gz`

I hope this explanation makes the regex clearer! Let me know if you have any other questions

augustin-v:
> OK. And the reason for being strict is to reduce the likelihood of false positives?

Yes, if it was looser there could be a case falsely flagging URLs for example.

> Would it also make sense to handle paths beginning with `./` (i.e., one dot)?

You are right! It does make sense, that is a great suggestion we'd just need to add it to the start of the regex `((?:\./|\.\./|`etc..
Would you like me to implement this change, or any other suggestions for improving the lint?

examples/supplementary/nonexistent_path_in_comment/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+crate-type = ["cdylib"]
+
+[dependencies]
+clippy_utils = { git = "https://github.com/rust-lang/rust-clippy", rev = "19e305bb57a7595f2a8d81f521c0dd8bf854e739" }
+dylint_linting = "3.5.1"
+regex = { workspace = true }
+once_cell = { workspace = true }
+
+[dev-dependencies]
+dylint_testing = "3.5.1"
```
smoelius, line 19:
```suggestion
dylint_testing = { path = "../../../utils/testing" }
```

examples/supplementary/nonexistent_path_in_comment/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// fn main() {}
+    /// ```
+    /// Use instead:
+    /// ```
+    /// // See ../actual/path/file.rs for implementation details
+    /// fn main() {}
+    /// ```
+    pub NONEXISTENT_PATH_IN_COMMENT,
+    Warn,
+    "reports file paths in comments that do not exist"
```
smoelius, line 34:
```suggestion
    "file paths in comments that do not exist"
```
For consision.

examples/supplementary/nonexistent_path_in_comment/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+static BLOCK_COMMENT: Lazy<Regex> = Lazy::new(|| Regex::new(r"/\*(([^*]|\*[^/])*)\*/").unwrap());
+static PATH_REGEX: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"((?:\.\./|/|[\w/-]+/)+[\w-]+(?:\.[\w-]+)+)").unwrap());
+
+impl<'tcx> LateLintPass<'tcx> for NonexistentPathInComment {
+    fn check_crate(&mut self, cx: &LateContext<'tcx>) {
+        let source_map = cx.tcx.sess.source_map();
+
+        for file in source_map.files().iter() {
+            if let Ok(content) = file.src.as_ref().ok_or("err") {
```
smoelius, line 47:
```suggestion
            if let Some(content) = file.src.as_ref() {
```

examples/supplementary/nonexistent_path_in_comment/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// ```
+    pub NONEXISTENT_PATH_IN_COMMENT,
+    Warn,
+    "reports file paths in comments that do not exist"
+}
+
+static LINE_COMMENT: Lazy<Regex> = Lazy::new(|| Regex::new("(^|[^/])(//([^/].*))").unwrap());
+static BLOCK_COMMENT: Lazy<Regex> = Lazy::new(|| Regex::new(r"/\*(([^*]|\*[^/])*)\*/").unwrap());
+static PATH_REGEX: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"((?:\.\./|/|[\w/-]+/)+[\w-]+(?:\.[\w-]+)+)").unwrap());
```
smoelius, line 40: Please update this regex to handle paths starting with a single dot (i.e., `./`).

examples/supplementary/nonexistent_path_in_comment/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    for captures in PATH_REGEX.captures_iter(comment_text) {
+        let path_str = &captures[1];
+        let full_path = base_dir.join(path_str);
+
+        if !full_path.exists() {
+            span_lint_and_help(
+                cx,
+                NONEXISTENT_PATH_IN_COMMENT,
+                span,
+                format!("referenced path does not exist: {}", path_str),
```
smoelius, line 109:
```suggestion
                "referenced path does not exist",
```
This will be a non-trivial change, but do you think it would be possible to have the span point directly to the matched path?

augustin-v:
Hi @smoelius, I applied the requested changes and narrowed down the span to point at only the captured path. I did it by changing the start of the span to be at the substring captured by the path regex.

Do you think it would be possible to make it support a pattern where paths are relative to the project root, like in the initial issue that you opened?
At first I thought about looking upwards to find a `Cargo.toml` but I realized that there being a `Cargo.toml` file does not necessarily mean that it is the project root; maybe looking for the `.git` would work?

smoelius:
> Hi @smoelius, I applied the requested changes and narrowed down the span to point at only the captured path. I did it by changing the start of the span to be at the substring captured by the path regex.

Thanks!

> Do you think it would be possible to make it support a pattern where paths are relative to the project root, like in the initial issue that you opened? At first I thought about looking upwards to find a `Cargo.toml` but I realized that there being a `Cargo.toml` file does not necessarily mean that it is the project root; maybe looking for the `.git` would work?

I tried to do something like this in `escaping_doc_link`: https://github.com/trailofbits/dylint/blob/097161c2ab885bd6c70dc16c89719228d4e54d1d/examples/supplementary/escaping_doc_link/src/lib.rs#L48-L64

My idea was to:
- get the directory containing the source file being analyzed
- call `cargo metadata` in that directory
- find the package containing that directory

I think that approach could work here.

Please note that I haven't looked at the code I cited in a while, and I apologize if it contains bugs.

augustin-v:
> My idea was to:
> 
> * get the directory containing the source file being analyzed
> * call `cargo metadata` in that directory
> * find the package containing that directory

Thank you for sharing your approach! Using cargo metadata with the source file sounds promising indeed. I'll experiment with this idea and see how well I can integrate it. Appreciate the guidance, I'll keep you updated on the progress

smoelius:
> Hi @smoelius, hope you had a good weekend.

I did. I hope you did as well. :)

> Over the weekend I experimented with handling workspace roots and managed to get it working for the workspace root detection. However, I noticed that the tool still doesn't identify the dylint project itself as the workspace root. You can verify this by running:
> 
> ```
> cargo locate-project --workspace
> ```
> 
> On my setup, it returns:
> 
> ```
> {"root":"/Users/user/dylint/examples/supplementary/Cargo.toml"}
> ```

That makes sense actually. The repo is currently organized as several workspaces, one of which is `supplementary`: https://github.com/trailofbits/dylint/blob/e89f5b690fb6f4f1aea285562c1fa8e4acc082ed/examples/supplementary/Cargo.toml#L33

If lint is able to find `supplementary` as the workspace root, I think that's fine.

Thank you very much for working on this. I will finish my review within the next few days.

examples/README.md
```diff
@@ -79,4 +80,3 @@ The example libraries are separated into the following three categories:
 [general]: #general
 [restriction]: #restriction
 [supplementary]: #supplementary
-[testing]: #testing
```
smoelius, line 82: Mistake?

smoelius (commented):
This looks fantastic. I noticed one small thing that looks like a typo. Additionally, the current CI failure is caused by the README and doc comments being out of sync. If you run the following command, it will cause the README to match the doc comments:
```
BLESS=1 cargo test format_example_readmes
```
So, if you could, please get the doc comments to how you would like them, and then run that command.

examples/README.md
augustin-v: Yes! I thought that I had to remove it at the time, because I had forgotten to add a new line

smoelius:
Thanks a lot!

smoelius:
Ugh. I think the merge commits may be causing problems. Could I trouble you to rebase?

smoelius:
Actually, it looks like I was able to rebase it.

augustin-v:
> Actually, it looks like I was able to rebase it.

Sorry, I should have been the one to do it. Thank you very much!

Issue #1533, opened 2025-02-19
Need a way to verify that `register_lints` registers all workspace members


smoelius:
E.g., that this `register_lints` function registers all the `supplementary` workspace members: https://github.com/trailofbits/dylint/blob/5cb7877f3ecb691e4fe6bdf3c8f2d2b123653ca4/examples/supplementary/src/lib.rs#L11-L20

At the moment, I'm not sure how best to do this.

augustin-v:
I'd love to work on this! I'm guessing this is a function that would go to `cargo-dylint/tests/ci.rs` ?

smoelius:
Yeah.

Do you have thoughts on how you would implement this?

The best idea I have come up with is:
- Find each example library workspace ([examples::iter](https://github.com/trailofbits/dylint/blob/0977ac58bbff20c1aacab3e7cc68a5576c301cc2/internal/src/examples.rs#L30-L36) might help with that).
- Then, for each:
  - Extract the `register_lints` calls from the root package's src/lib.rs file using regular expressions.
  - Get the workspace's packages using `cargo metadata`.
  - Compare the extracted `register_lints` calls to the list of found packages.

I'm open to other ideas though.

Please be aware that something in Dylint broke with the upgrade to Rust 1.85. I need to investigate.

augustin-v:
That sounds very good and this made me realize that I forgot to add the `nonexistent_path_in_comment` to the lib.rs file.

> * Find each example library workspace ([examples::iter](https://github.com/trailofbits/dylint/blob/0977ac58bbff20c1aacab3e7cc68a5576c301cc2/internal/src/examples.rs#L30-L36) might help with that).

This seems perfect for the task indeed, I was thinking about how we could fetch the examples with multiple workspaces before looking at it, and that does just that.

Definitely doable! Could I work on it?

smoelius:
> Definitely doable! Could I work on it?

Yes! Please!

smoelius:
Fixed by #1535

PR #1535, opened 2025-02-21
feat#1533: verify  `general` and `supplementary`  registered lints

augustin-v:
Relevant issue #1533 

This test ensures that all lints defined in the "general" and "supplementary" folders are actually registered in the corresponding src/lib.rs file. The process is as follows:
1. Extract expected lints by listing the immediate subdirectories in "general" and "supplementary" (ignoring src and hidden directories) to determine the expected lint module names.
2. Read the src/lib.rs file in both folder and extract registered lints using this regex 
`([a-zA-Z_][a-zA-Z0-9_]*)::register_lints\s*\(` which captures the module name right before the literal `::register_lints` call.
4. Lastly, compare the expected and actually registered lints.

if we were to run the test with the dylint's current state we would get this output:
```bash
---- verify_registered_lints stdout ----

thread 'verify_registered_lints' panicked at cargo-dylint/tests/ci.rs:584:13:
Mismatch in /Users/user/dylint/cargo-dylint/../examples/supplementary:

Missing registered lints: ["local_ref_cell", "nonexistent_path_in_comment"]

Expected: ["commented_code", "escaping_doc_link", "inconsistent_struct_pattern", "local_ref_cell", "nonexistent_path_in_comment", "redundant_reference", "unnamed_constant", "unnecessary_borrow_mut", "unnecessary_conversion_for_trait"]
Actual: ["commented_code", "escaping_doc_link", "inconsistent_struct_pattern", "redundant_reference", "unnamed_constant", "unnecessary_borrow_mut", "unnecessary_conversion_for_trait"]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Also seeing that `::register_lints` is only used inside of the general and supplementary folders, I figured that using the examples::iter function would be too broad for our check.

This approach is a bit different from the initial that we initially discussed about, because we extract the existing lints with their folder names instead of extracting them from the `cargo metadata`. I'm open to feedback or alternative suggestions!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -939,6 +939,7 @@ dependencies = [
  "cargo-util",
  "cargo_metadata",
  "dylint_internal",
+ "regex",
  "toml",
  "toml_edit",
  "walkdir",
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -9,15 +9,15 @@ publish = false
 [dev-dependencies]
 cargo-util = { workspace = true }
 cargo_metadata = { workspace = true }
+regex = { workspace = true }
+toml = { workspace = true }
+toml_edit = { workspace = true }
+walkdir = { workspace = true }
 
 dylint_internal = { version = "=4.1.0", path = "../internal", features = [
     "clippy_utils",
     "examples",
 ] }
-regex = { workspace = true }
-toml = { workspace = true }
-toml_edit = { workspace = true }
-walkdir = { workspace = true }
 
 [lints]
 workspace = true
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -9,14 +9,15 @@ publish = false
 [dev-dependencies]
 cargo-util = { workspace = true }
 cargo_metadata = { workspace = true }
-toml = { workspace = true }
-toml_edit = { workspace = true }
-walkdir = { workspace = true }
 
 dylint_internal = { version = "=4.1.0", path = "../internal", features = [
     "clippy_utils",
     "examples",
 ] }
+regex = { workspace = true }
+toml = { workspace = true }
+toml_edit = { workspace = true }
+walkdir = { workspace = true }
 
 [lints]
 workspace = true
--- a/examples/src/lib.rs
+++ b/examples/src/lib.rs
@@ -4,7 +4,14 @@ mod test {
     use dylint_internal::{
         CommandExt, clippy_utils::toolchain_channel, examples::iter, rustup::SanitizeEnvironment,
     };
-    use std::{ffi::OsStr, fs::read_to_string, process::Command};
+    use regex::Regex;
+    use std::{
+        collections::BTreeSet,
+        ffi::OsStr,
+        fs::{read_dir, read_to_string},
+        path::Path,
+        process::Command,
+    };
     use toml_edit::{DocumentMut, Item};
     use walkdir::WalkDir;
 
@@ -221,5 +228,57 @@ mod test {
                 .collect::<Vec<_>>()
                 .join("\n")
         );
+    }
+
+    #[test]
+    fn verify_registered_lints() {
+        const CATEGORIES: &[&str] = &["general", "supplementary"];
+        let register_lints_re =
+            Regex::new(r"([a-zA-Z_][a-zA-Z0-9_]*)::register_lints\s*\(").unwrap();
+
+        for category in CATEGORIES {
+            let category_path = Path::new(category);
+            let lib_rs_path = category_path.join("src/lib.rs");
+            let file_contents = read_to_string(&lib_rs_path)
+                .unwrap_or_else(|e| panic!("Failed to read {}: {}", lib_rs_path.display(), e));
+
+            let actual_lints: BTreeSet<_> = register_lints_re
+                .captures_iter(&file_contents)
+                .map(|cap| cap.get(1).unwrap().as_str().to_string())
+                .collect();
+
+            let expected_lints: BTreeSet<_> = read_dir(category_path)
+                .unwrap_or_else(|e| {
+                    panic!(
+                        "Failed to read directory {}: {}",
+                        category_path.display(),
+                        e
+                    )
+                })
+                .filter_map(|entry| {
+                    let entry = entry.ok()?;
+                    let path = entry.path();
+                    if path.is_dir() {
+                        let file_name = path.file_name()?;
+                        let dir_name = file_name.to_str()?;
+                        if dir_name != "src" && !dir_name.starts_with('.') {
+                            return Some(dir_name.to_string());
+                        }
+                    }
+                    None
+                })
+                .collect();
+
+            let missing: Vec<_> = expected_lints.difference(&actual_lints).collect();
+
+            assert!(
+                missing.is_empty(),
+                "Mismatch in {}\n\nMissing registered lints: {:?}\n\nExpected: {:?}\nActual: {:?}",
+                category_path.display(),
+                missing,
+                expected_lints,
+                actual_lints
+            );
+        }
     }
 }


smoelius:
I really appreciate you working on this. 🙏

> ```shell
> Missing registered lints: ["local_ref_cell", "nonexistent_path_in_comment"]
> ```

I noticed this actually. :)

I started a PR to run the `nonexistent_path_in_comment` lint over the Dylint repo (#1536), and it is currently failing. I think the path regex may need to be refined.

_You are welcome to ignore the failures_, but if you have ideas on how to refine the regex, I would welcome them.

Please note that I already made some small changes to the lint's logic: https://github.com/trailofbits/dylint/pull/1536/files#diff-655df697dba7900d7f7a61d9548c870a5068b918eabf6381760f107b35afb3a2L133-L149

Needless to say, I really appreciate you building that lint, because I am sure it will catch mistakes.

I will give the present PR a more thorough review after #1536 is finalized and merged.

augustin-v:
> I really appreciate you working on this. 🙏
Pleasure is mine, I'll be here for a while😊 

Not a very creative idea but maybe we could change the path regex to 
`(?:(?!\$)(?:\./|\.\./|/|[\w/-]+/)+[\w-]+(?:\.[\w-]+)+)` so that it filters out paths starting with `$` and add when a comment is flagged, we would add a note to make it user friendly such as : 
---- "NOTE: add $ prefix at the start of the path to ignore" ----

smoelius:
> Pleasure is mine I'll be here for a while, I have a lot to learn from you😊

You're very kind. :)

> Not a very creative idea but maybe we could change the path regex to `(?:(?!\$)(?:\./|\.\./|/|[\w/-]+/)+[\w-]+(?:\.[\w-]+)+)` so that it filters out paths starting with `$` and add when a comment is flagged, we would add a note to make it user friendly such as : ---- "NOTE: add $ prefix at the start of the path to ignore" ----

It might be worth just `allow`ing the `$DIR/...` cases.

But I happen to know the lint triggers in other places. For example, it produces these warnings in the `general` workspace:
```
warning: referenced path does not exist
  --> local_ref_cell/src/lib.rs:37:44
   |
37 |     /// [`RefCell`]: https://doc.rust-lang.org/std/cell/struct.RefCell.html
   |                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: verify the path is correct or remove the reference
   = note: `#[warn(nonexistent_path_in_comment)]` on by default

warning: `local_ref_cell` (lib) generated 1 warning
warning: referenced path does not exist
  --> unnecessary_borrow_mut/src/lib.rs:54:56
   |
54 |     /// [`RefCell::borrow_mut`]: https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.borrow_mut
   |                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: verify the path is correct or remove the reference
   = note: `#[warn(nonexistent_path_in_comment)]` on by default

warning: referenced path does not exist
  --> unnecessary_borrow_mut/src/lib.rs:55:52
   |
55 |     /// [`RefCell::borrow`]: https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.borrow
   |                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: verify the path is correct or remove the reference
```

Another idea, besides refining the regex, could be to cast a wider net and then filter out unwanted matches afterward.

Like maybe the regex could match non-empty sequences with at least one `/`, and then the lint does:
```rs
if full_path.starts_with("https://") {
    continue;
}
```

augustin-v:
> Another idea, besides refining the regex, could be to cast a wider net and then filter out unwanted matches afterward.
> 
> Like maybe the regex could match non-empty sequences with at least one `/`, and then the lint does:
> 
> ```rust
> if full_path.starts_with("https://") {
>     continue;
> }
> ```

Yes this seems like I the best approach to deal with the url similar to the escaping_doc_link one I guess; looks like I did not implement the lint to ignore URLs as well as I thought 🤔

smoelius:
Doing what I said literally produces a lot of false positives:
```
warning: referenced path does not exist
  --> internal/src/rustup.rs:46:41
   |
46 |     // smoelius: `path` should end with `/bin/rustc`.
   |                                         ^^^^^^^^^^^^^
   |
   = help: verify the path is correct or remove the reference
   = note: `#[warn(nonexistent_path_in_comment)]` on by default

warning: `dylint_internal` (lib) generated 1 warning
warning: referenced path does not exist
  --> dylint/src/lib.rs:22:26
   |
22 | // smoelius: See note in dylint/src/library_packages/mod.rs.
   |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: verify the path is correct or remove the reference
   = note: `#[warn(nonexistent_path_in_comment)]` on by default

warning: referenced path does not exist
  --> dylint/src/driver_builder.rs:21:10
   |
21 | (https://github.com/trailofbits/dylint).
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: verify the path is correct or remove the reference

warning: referenced path does not exist
   --> dylint/src/driver_builder.rs:159:13
    |
159 |     // like `$ORIGIN/../../`... (see https://github.com/trailofbits/dylint/issues/54). The new
    |             ^^^^^^^^^^^^^^^^^^^
    |
    = help: verify the path is correct or remove the reference
```
I'm struggling a little bit here. I'm not sure what the best path forward is.

cargo-dylint/tests/ci.rs
```diff
@@ -534,6 +535,63 @@ fn update() {
     });
 }
 
+#[test]
+fn verify_registered_lints() {
+    let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
+    let examples_dir = manifest_dir.join("../examples");
+
+    let folders = ["general", "supplementary"];
```
smoelius, line 543:
```suggestion
    const CATEGORIES: &[&str] = &["general", "supplementary"];
```
Nit. I think "categories" is more descriptive than "folders". (Also, I think "folder" is technically a Windows term.)

cargo-dylint/tests/ci.rs
```diff
@@ -534,6 +535,63 @@ fn update() {
...
+
+    let register_lints_re =
+        Regex::new(r"([a-zA-Z_][a-zA-Z0-9_]*)::register_lints\s*\(").expect("Invalid regex");
+
+    for folder in folders.iter() {
```
smoelius, line 548:
```suggestion
    for category in CATEGORIES {
```

cargo-dylint/tests/ci.rs
```diff
@@ -534,6 +535,63 @@ fn update() {
     });
 }
 
+#[test]
+fn verify_registered_lints() {
+    let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
+    let examples_dir = manifest_dir.join("../examples");
```
smoelius, line 541:
```suggestion
```
After the test is moved, these should no longer be necessary.

cargo-dylint/tests/ci.rs
```diff
@@ -534,6 +535,63 @@ fn update() {
...
+
+        if !missing.is_empty() {
+            panic!(
+                "Mismatch in {}:\n\nMissing registered lints: {:?}\n\nExpected: {:?}\nActual: {:?}",
+                folder_path.display(),
+                missing,
+                expected_lints,
+                extracted_lints
+            );
+        }
```
smoelius, line 591: I think this can be rewritten as an `assert`.

cargo-dylint/tests/ci.rs
```diff
@@ -534,6 +535,63 @@ fn update() {
...
+        Regex::new(r"([a-zA-Z_][a-zA-Z0-9_]*)::register_lints\s*\(").expect("Invalid regex");
+
+    for folder in folders.iter() {
+        let folder_path = examples_dir.join(folder);
+
+        let lib_rs_path = folder_path.join("src").join("lib.rs");
+        let file_contents = read_to_string(&lib_rs_path)
+            .unwrap_or_else(|e| panic!("Failed to read {}: {}", lib_rs_path.display(), e));
+
+        let mut extracted_lints: Vec<String> = register_lints_re
```
smoelius, line 555:
```suggestion
        let mut actual_lints: Vec<String> = register_lints_re
```
Nit. To adhere to the convention of comparing "expected" and "actual".

cargo-dylint/tests/ci.rs
```diff
@@ -534,6 +535,63 @@ fn update() {
...
+                    let dir_name = path.file_name()?.to_str()?;
+                    if dir_name != "src" && !dir_name.starts_with('.') {
+                        return Some(dir_name.to_string());
+                    }
+                }
+                None
+            })
+            .collect();
+        expected_lints.sort();
+        expected_lints.dedup();
```
smoelius, line 577:
```suggestion
```
Same comment as above.

cargo-dylint/tests/ci.rs
```diff
@@ -534,6 +535,63 @@ fn update() {
...
+        let lib_rs_path = folder_path.join("src").join("lib.rs");
+        let file_contents = read_to_string(&lib_rs_path)
+            .unwrap_or_else(|e| panic!("Failed to read {}: {}", lib_rs_path.display(), e));
+
+        let mut extracted_lints: Vec<String> = register_lints_re
+            .captures_iter(&file_contents)
+            .map(|cap| cap.get(1).unwrap().as_str().to_string())
+            .collect();
+        extracted_lints.sort();
+        extracted_lints.dedup();
```
smoelius, line 560:
```suggestion
```
If the lints are going to be put into a `BTreeSet`, there should be no need to sort them.

cargo-dylint/tests/ci.rs
```diff
@@ -534,6 +535,63 @@ fn update() {
...
+
+        let mut expected_lints: Vec<String> = read_dir(&folder_path)
+            .unwrap_or_else(|e| panic!("Failed to read directory {}: {}", folder_path.display(), e))
+            .filter_map(|entry| {
+                let entry = entry.ok()?;
+                let path = entry.path();
+                if path.is_dir() {
+                    let dir_name = path.file_name()?.to_str()?;
```
smoelius, line 568: The Dylint lint `question_mark_in_expression` should fire on this code.

But even if it doesn't, could you rewrite this code so that the question marks are the outermost operators in the expressions in which they appear?

It's a personal preference, but I dislike having an operator that can hijack control flow in the middle of an expression.

cargo-dylint/tests/ci.rs
```diff
@@ -534,6 +535,63 @@ fn update() {
...
+                    }
+                }
+                None
+            })
+            .collect();
+        expected_lints.sort();
+        expected_lints.dedup();
+
+        let expected_set: BTreeSet<_> = expected_lints.iter().cloned().collect();
+        let extracted_set: BTreeSet<_> = extracted_lints.iter().cloned().collect();
```
smoelius, line 580:
```suggestion
        let actual_set: BTreeSet<_> = extracted_lints.iter().cloned().collect();
```
Again, to adhere to conventions.

smoelius (commented):
@augustin-v Thanks very much for your patience, and thanks very much for preparing this test.

I know I said this test should go in ci.rs. But now I realize it would be better placed in examples/src/lib.rs. Would it be possible to move the test there?

I may need to give this another look after you've had a chance to review my comments. I hope you don't mind.

smoelius:
@augustin-v Gentle ping. If you are busy, I would be happy to push this over the finish line.

augustin-v:
> @augustin-v Gentle ping. If you are busy, I would be happy to push this over the finish line.

Apologies for the very late reaction. Thank you for the ping, working on it now

smoelius:
> Apologies for the very late reaction. Thank you for the ping, working on it now

No worries at all. Thank you very much!

augustin-v:
Hi @smoelius, I've addressed the changes please tell me if I forgot anything or anything you'd like me to change in the code, I'll fix it right away.

>But even if it doesn't, could you rewrite this code so that the question marks are the outermost operators in the expressions in which they appear?

I found this very interesting! never thought about it but it makes sense for debugging and readability

smoelius:
Thanks, @augustin-v. I will look at this within the next few days.

smoelius:
Thanks a lot, @augustin-v.

augustin-v:
The pleasure is mine @smoelius thanks for the help on the CI. I just got back from a long trip, looking forward to contribute again!

PR #1536, opened 2025-02-21
Register `nonexistent_path_in_comment` in `supplementary` library

smoelius:

PR #1540, opened 2025-02-24
Fix npic false positive

augustin-v:
Hi @smoelius! I've researched on my own about how to filter out the false positives we are encountering(#1536) consisting of:
- Complex URL schemas
- Paths for demonstration, starting with `$`

I managed to make significant progress, especially with filtering out these two.
The first notable change from the lint is the switch to the `fancy-regex` crate, which allow more flexibility by using look-around assertions in regular expressions. Here's a detailed breakdown of the key improvements:

1. URL filtering with  `fancy-regex`
- Added `URL_REGEX` pattern `https?://\S+` to detect web URLs
- Implemented URL range exclusion logic to avoid path checks in URLs

2. Improved path pattern matching
- Updated `PATH_REGEX` to support:
  - Dollar signs in paths (`\$`)
- Added explicit exclusion for paths starting with:
  - URL protocols (`http://`, `https://`)
  - Web domains (`www.`)
  - Placeholder indicators (`$`)

These changes significantly reduce false positives while maintaining strong detection of actual broken path references. The fancy-regex crate enables complex pattern matching that would panic with the standard regex crate, particularly the negative lookbehind needed for accurate URL detection within comments.

The test suite has also been updated to verify these new cases, including:
- URLs containing path-like fragments
- Documentation comments with intentional examples
- Placeholder paths using `$DIR/path` syntax

Looking forward to your feedback!


smoelius:
Sorry, I've just been busy with other things. I will review this within the next few days.

smoelius:
> These changes significantly reduce false positives while maintaining strong detection of actual broken path references. The fancy-regex crate enables complex pattern matching that would panic with the standard regex crate, particularly the negative lookbehind needed for accurate URL detection within comments.

Thanks a lot for working on this. I'm not familiar with `fancy-regex`, but [its docs](https://docs.rs/fancy-regex/latest/fancy_regex/) suggest that lookbehind would use a syntax like `(?<=exp)` or `(?<!exp)`. I don't see either of those in the code. Am I making a mistake by looking for them?

More generally, I'm have a little trouble with the logic. Could you explain what `url_ranges` and `non_url_ranges` are, intuitively?

smoelius:
@augustin-v Thanks again for working on this.

I will get to reviewing #1535 soon.

PR #1546, opened 2025-03-03
Account for new `rustup show active-toolchain` format

smoelius:


TimDiekmann:
Out of interest: if I set `RUSTUP_AUTO_INSTALL=0` for _a good reason_ (I don't want any tool to alter my environment), why should `cargo-dylint` bypass this security consideration?

indietyp:
I echo the sentiment of @TimDiekmann, I would expect `cargo-dylint` to error out gracefully with a message similar to:

"Toolchain <name> wasn't found, and auto-install was disabled via `RUSTUP_AUTO_INSTALL=0`, specify `--install-toolchain` for `cargo dylint` to automatically manage and download the required toolchain or execute `rustup toolchain install <name>` to install the toolchain manually."

smoelius:
> Out of interest: if I set `RUSTUP_AUTO_INSTALL=0` for _a good reason_ (I don't want any tool to alter my environment), why should `cargo-dylint` bypass this security consideration?

Currently, if you set `RUSTUP_AUTO_INSTALL=0` and try to run `cargo-dylint`, things break.

I can't remember precisely where. The bottom line is: `cargo-dylint` expects toolchains to be automatically installed when they are not.

So the intent is not to bypass the security restriction, but rather to work within its bounds.

Parhaps I need to rephrase the issue (or have I misunderstood your question)?

smoelius:
> I echo the sentiment of @TimDiekmann, I would expect `cargo-dylint` to error out gracefully with a message similar to:
> 
> "Toolchain wasn't found, and auto-install was disabled via `RUSTUP_AUTO_INSTALL=0`, specify `--install-toolchain` for `cargo dylint` to automatically manage and download the required toolchain or execute `rustup toolchain install <name>` to install the toolchain manually."

🤔 Thank you for the feedback. I will try to give this some more thought.

TimDiekmann:
The issue is that if you set `RUSTUP_AUTO_INSTALL=0`, you likely do that for a very good reason (it's not the default since rustup 1.28.1). You do that so no toolchain is installed automatically. 

> `cargo-dylint` expects toolchains to be automatically installed when they are not.

`cargo-dylint` expects the toolchain to be installed and invocations to `rustup` without `RUSTUP_AUTO_INSTALL=0` (or with `RUSTUP_AUTO_INSTALL=1`, which is the default) will automatically install the toolchain. But if you set `RUSTUP_AUTO_INSTALL=0` you explicitly opt-out of that behavior.

As an alternative, it might be interesting to provide a human readable error, such as "Run `rustup toolchain install nightly-2024-05-14`", as mentioned by @indietyp. It seems to be possible to read the list of toolchains currently by running `cargo dylint list`, so it would be possible to also use that output to generate a batch of commands to install required toolchain.

smoelius:
@TimDiekmann I think I agree with you.

Thank both you and @indietyp. You've caused me to look at this in a way I hadn't before. 🙏

TimDiekmann:
Great, thanks of taking a look and glad I could help 🙂

smoelius:
I'm closing this. When I opened it, `RUSTUP_AUTO_INSTALL=0` was the default. Now that it's not, I'm not sure what's left here to do.

Thanks very much to @indietyp and @TimDiekmann for their comments.

Issue #636, opened 2023-03-12
Convert GitHub workflow tests to Rust tests


smoelius:
Do more with Rust tests and less with Yaml/Bash.

The point is to make more tests [runnable via `cargo test`](https://www.reddit.com/r/rust/comments/prhz1u/comment/hdkbpjl/) (not to use Rust for Rust's sake).

- [x] [Prettier](https://github.com/trailofbits/dylint/blob/178b740594bc97968adf68f0d8a139c46f807455/.github/workflows/ci.yml#L37)
- [x] [Cargo sort](https://github.com/trailofbits/dylint/blob/178b740594bc97968adf68f0d8a139c46f807455/.github/workflows/ci.yml#L42)
- [x] [Format](https://github.com/trailofbits/dylint/blob/178b740594bc97968adf68f0d8a139c46f807455/.github/workflows/ci.yml#L47)
- [x] [Format example READMEs](https://github.com/trailofbits/dylint/blob/178b740594bc97968adf68f0d8a139c46f807455/.github/workflows/ci.yml#L55)
  - [x] Main README
  - [x] Individual READMEs
- [x] [Format util READMEs](https://github.com/trailofbits/dylint/blob/178b740594bc97968adf68f0d8a139c46f807455/.github/workflows/ci.yml#L58)
- [x] [Check lockfiles](https://github.com/trailofbits/dylint/blob/178b740594bc97968adf68f0d8a139c46f807455/.github/workflows/ci.yml#L63)
- [x] [Check licenses](https://github.com/trailofbits/dylint/blob/178b740594bc97968adf68f0d8a139c46f807455/.github/workflows/ci.yml#L66)
- [x] `shellcheck`: https://github.com/trailofbits/dylint/blob/7fd2c4d410eb0b7744f26c224e1eb3c23083e551/.github/workflows/ci.yml#L78
- [ ] lint.sh script: https://github.com/trailofbits/dylint/blob/7fd2c4d410eb0b7744f26c224e1eb3c23083e551/.github/workflows/ci.yml#L111

Necessist uses this approach, e.g., for [checking licenses](https://github.com/trailofbits/necessist/blob/7d46abec83529e449a6b7fbc7bacf73b02e1079c/core/tests/ci.rs#L65-L84) and [checking lockfiles](https://github.com/trailofbits/necessist/blob/7d46abec83529e449a6b7fbc7bacf73b02e1079c/core/tests/ci.rs#L170-L178).

[...]

PR #1563, opened 2025-03-25
Convert README verification script to a Rust test

jackcooper20:
This PR addresses #636 by replacing the bash script (`scripts/update_example_READMEs.sh`) with a Rust-based test that verifies the examples/README.md contents are in sync with the actual examples in the codebase.

The test:
- Walks through all example categories (general, supplementary, restriction, experimental, and testing)
- Extracts example names and descriptions from each Cargo.toml file
- Generates a temporary README.md with the expected content
- Verifies that the actual README.md contains all necessary example links and descriptions

This change makes the README verification process part of the standard test suite, runnable via `cargo test`. By keeping this verification in Rust rather than bash, we ensure better cross-platform compatibility and integration with the rest of the testing infrastructure.

### DIFF
--- a/cargo-dylint/tests/ci.rs
+++ b/cargo-dylint/tests/ci.rs
@@ -19,6 +19,9 @@ use std::{
 
 static METADATA: LazyLock<Metadata> = LazyLock::new(|| current_metadata().unwrap());
 
+static DESCRIPTION_REGEX: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r#"description\s*=\s*"([^"]*)""#).unwrap());
+
 #[ctor::ctor]
 fn initialize() {
     set_current_dir("..").unwrap();
@@ -131,6 +134,168 @@ fn cargo_dylint_and_dylint_readmes_are_equal() {
     let dylint_readme = readme_contents("dylint").unwrap();
 
     compare_lines(&cargo_dylint_readme, &dylint_readme);
+}
+
+#[test]
+fn examples_readme_contents() {
+    let examples_dir = Path::new(env!("CARGO_MANIFEST_DIR")).join("../examples");
+    let categories = vec![
+        "general",
+        "supplementary",
+        "restriction",
+        "experimental",
+        "testing",
+    ];
+
+    // Read the existing README
+    let readme_path = examples_dir.join("README.md");
+    let readme_content = read_to_string(&readme_path).unwrap();
+
+    // Generate just the lint description tables
+    let expected_tables = generate_lint_tables(&examples_dir, &categories);
+
+    // Extract the current tables section from README using markers
+    let actual_tables = extract_between_markers(&readme_content)
+        .unwrap_or_else(|| panic!("Lint description markers not found in README.md"));
+
+    // Compare the trimmed generated content with the actual content
+    assert_eq!(
+        expected_tables.trim(),
+        actual_tables,
+        "Lint descriptions in README.md do not match expected content"
+    );
+}
+
+fn generate_lint_tables(examples_dir: &Path, categories: &[&str]) -> String {
+    const EXAMPLE_HEADER: &str = "Example";
+    const DESC_HEADER: &str = "Description/check";
+
+    let mut content = String::new();
+
+    // Generate the tables for each category
+    for category in categories {
+        use std::cmp::max;
+        use std::fmt::Write;
+
+        // Get the examples for this category
+        let examples = collect_examples_from_category(examples_dir, category);
+
+        // Calculate column widths
+        let max_example_width = examples
+            .iter()
+            .map(|(name, _)| format!("[`{name}`](./{category}/{name})").len())
+            .max()
+            .unwrap_or(0);
+
+        let max_desc_width = examples
+            .iter()
+            .map(|(_, description)| description.len())
+            .max()
+            .unwrap_or(0);
+
+        let example_col_width = max(EXAMPLE_HEADER.len(), max_example_width);
+        let desc_col_width = max(DESC_HEADER.len(), max_desc_width);
+
+        // Write header
+        write!(content, "\n## {}\n\n", capitalize(category)).unwrap();
+        #[allow(clippy::uninlined_format_args)]
+        writeln!(
+            content,
+            "| {:<example_col_width$} | {:<desc_col_width$} |",
+            EXAMPLE_HEADER, DESC_HEADER
+        )
+        .unwrap();
+        writeln!(
+            content,
+            "| {:-<example_col_width$} | {:-<desc_col_width$} |",
+            "", ""
+        )
+        .unwrap(); // Separator line
+
+        // Write rows with padding
+        for (name, description) in examples {
+            let example_link = format!("[`{name}`](./{category}/{name})");
+            #[allow(clippy::uninlined_format_args)]
+            writeln!(
+                content,
+                "| {:<example_col_width$} | {:<desc_col_width$} |",
+                example_link, description
+            )
+            .unwrap();
+        }
+    }
+
+    content
+}
+
+fn extract_between_markers(content: &str) -> Option<String> {
+    const START_MARKER: &str = "<!-- lint descriptions start -->";
+    const END_MARKER: &str = "<!-- lint descriptions end -->";
+    let start = content.find(START_MARKER)? + START_MARKER.len();
+    let end = content.find(END_MARKER)?;
+    Some(content[start..end].trim().to_string())
+}
+
+fn collect_examples_from_category(examples_dir: &Path, category: &str) -> Vec<(String, String)> {
+    let mut examples = Vec::new();
+    let category_dir = examples_dir.join(category);
+
+    for entry in read_dir(&category_dir).unwrap() {
+        let entry = entry.unwrap();
+        let metadata = entry.metadata().unwrap();
+        if metadata.is_dir() {
+            let cargo_toml_path = entry.path().join("Cargo.toml");
+            if cargo_toml_path.exists() {
+                if let Some((name, desc)) = extract_name_and_description(&cargo_toml_path) {
+                    examples.push((name, desc));
+                }
+            }
+        }
+    }
+
+    // Sort examples by name
+    examples.sort_by(|(a, _), (b, _)| a.cmp(b));
+    examples
+}
+
+fn extract_name_and_description(cargo_toml_path: &Path) -> Option<(String, String)> {
+    let content = read_to_string(cargo_toml_path).ok()?;
+
+    // Get the name from the directory
+    let name = cargo_toml_path
+        .parent()
+        .and_then(|path| path.file_name())
+        .unwrap()
+        .to_string_lossy()
+        .to_string();
+
+    // Extract the description using regex
+    let description = if let Some(caps) = DESCRIPTION_REGEX.captures(&content) {
+        let desc = caps.get(1).unwrap();
+        // Format the description like the bash script does
+        let desc_str = desc.as_str();
+        if let Some(stripped) = desc_str.strip_prefix("A lint to check for ") {
+            capitalize(stripped)
+        } else {
+            desc_str.to_string()
+        }
+    } else {
+        return None;
+    };
+
+    Some((name, description))
+}
+
+fn capitalize(s: &str) -> String {
+    let mut chars = s.chars();
+    match chars.next() {
+        None => String::new(),
+        Some(first) => {
+            let mut result = first.to_uppercase().collect::<String>();
+            result.extend(chars);
+            result
+        }
+    }
 }
 
 #[cfg_attr(dylint_lib = "general", allow(non_thread_safe_call_in_test))]
--- a/examples/README.md
+++ b/examples/README.md
@@ -8,6 +8,8 @@ The example libraries are separated into the following three categories:
 - [experimental] - not ready for primetime yet (similar to [Clippy]'s "nursery" category)
 - [testing] - used only for testing purposes
 
+<!-- lint descriptions start -->
+
 ## General
 
 | Example                                                                                  | Description/check                                                      |
@@ -65,6 +67,8 @@ The example libraries are separated into the following three categories:
 | ---------------------------------- | ------------------------------------------------------ |
 | [`clippy`](./testing/clippy)       | All of the Clippy lints as a Dylint library            |
 | [`straggler`](./testing/straggler) | A lint that uses an old toolchain for testing purposes |
+
+<!-- lint descriptions end -->
 
 **Notes**
 


jackcooper20:
hey @smoelius, here is the PR. Please check and let me know if any thing is needed. Thank You!

smoelius:
Hey, thanks @jackcooper20! I will try to look at this no later than this weekend.

jackcooper20:
@smoelius completed the CI

examples/tests/test_readme.rs
```diff
@@ -0,0 +1,242 @@
...
+
+#[allow(unknown_lints)]
+#[allow(inconsistent_qualification)]
+fn collect_examples_from_category(examples_dir: &Path, category: &str) -> Vec<(String, String)> {
+    let mut examples = Vec::new();
+    let category_dir = examples_dir.join(category);
+    
+    if category == "restriction" {
+        // Handle restriction directory differently since it seems to have directories with slashes in names
+        for entry in read_dir(&category_dir).unwrap() {
```
smoelius, line 136: Can't you use `read_dir` for both restriction and non-restriction cases?

examples/tests/test_readme.rs
```diff
@@ -0,0 +1,242 @@
...
+
+fn extract_name_and_description(cargo_path: &Path) -> Option<(String, String)> {
+    let mut content = String::new();
+    if let Ok(mut file) = File::open(cargo_path) {
+        if file.read_to_string(&mut content).is_err() {
+            return None;
+        }
+    } else {
+        return None;
+    }
```
smoelius, line 180: Could you use `fs::read_to_string` here? https://doc.rust-lang.org/std/fs/fn.read_to_string.html

examples/tests/test_readme.rs
```diff
@@ -0,0 +1,242 @@
...
+    for category in &categories {
+        verify_category_in_readme(&examples_dir, category, &actual_content);
+    }
+}
+
+fn generate_expected_readme(examples_dir: &Path, categories: &[&str]) -> String {
+    let mut content = String::new();
+    
+    // Add the header
+    content.push_str("# Example Dylint libraries\n\n");
```
smoelius, line 49: I don't think I like the idea of storing the entire README contents in the test. Because then the test will have to change whenever the README changes.

I think I would prefer that the test find where the lint descriptions should go and generate just that part, taking everything else from the existing README.

I may not be obvious, but the existing shell script worked this way: https://github.com/trailofbits/dylint/blob/349807f4370740538cac08127b2202d54be0472d/scripts/update_example_READMEs.sh#L24

But I would be open to making things even easier for the test. For example, the README could contain markers like `<!-- lint descriptions start -->`/`<!-- lint descriptions end -->` to help the test know where the lint descriptions should go.

smoelius (commented):
Thanks a lot for working on this, @jackcooper20.

My two main concerns right now are that the test encodes the entire README, and that the test does not simply compare the expected README with the actual README. The former is elaborated on below; the latter is more of a question, so if you could please explain. 🙏

examples/tests/test_readme.rs
```diff
@@ -0,0 +1,242 @@
+use std::fs::{File, read_dir};
+use std::io::{Read, Write};
+use std::path::{Path, PathBuf};
+use regex::Regex;
+use walkdir::WalkDir;
+use tempfile::NamedTempFile;
+
+#[test]
+fn test_readme_contents() {
+    let examples_dir = find_examples_dir();
```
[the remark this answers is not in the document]
jackcooper20: Hardcoding would be simpler and more explicit. The main reason I chose that approach was to ensure the test works regardless of where it's run

examples/tests/test_readme.rs
```diff
@@ -0,0 +1,242 @@
...
+        .and_then(|mut file| file.read_to_string(&mut actual_content))
+        .unwrap();
+    
+    // For debugging purposes, print the paths
+    eprintln!("Temp file path: {temp_file:?}");
+    eprintln!("Actual README path: {readme_path:?}");
+    
+    
+    // Instead we use a more flexible approach that allows for formatting differences
+    // while ensuring all examples are properly documented
```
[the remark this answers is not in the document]
jackcooper20: Basically to ensures the README is functionally correct while being resilient to formatting changes.

examples/tests/test_readme.rs
jackcooper20: Thanks for the suggestion. After looking at it again, I realized the current approach is more verbose than necessary. Using `fs::read_to_string` would make the code more concise and readable while maintaining the same functionality.

examples/tests/test_readme.rs
jackcooper20: Yeah you're right that hardcoding the entire README is brittle. I'll refactor the test to generate and verify only the table sections.

smoelius (commented):
> Do you think anything else needs to be changed other than the mentioned ones?

I honestly won't know until I see the code.

I really appreciate you working on this.

jackcooper20:
Hello @smoelius,

I've tried to fix the issues you mentioned in the review. However, after pushing the changes, one of the CI tests failed. I attempted to resolve the issue, but I encountered a potential bug in the `update_example_READMEs.sh` script.

In particular, the problem seems to be with [this line](https://github.com/trailofbits/dylint/blob/master/scripts/update_example_READMEs.sh#L34)
```bash
sed 's/| A lint to check for \([^|]*\) |$/| \u\1 |/'
```

The intent of this line is to capitalize text (e.g., changing `abc` to `Abc`). However, the `\u` syntax isn't supported in GNU `sed`, which results in an extra `u` being introduced before the text (e.g., `uLint description` instead of `Lint description`). This issue is likely causing a mismatch between the expected and actual content in `README.md`.

Could you please check if this is the root cause of the issue, or if I might be overlooking something? If it’s not the cause, I'd appreciate your guidance in resolving the CI failure.

jackcooper20:
@smoelius ,please take a look at this at your convenience

smoelius:
Hey, sorry for not replying, @jackcooper20. I will try to look at this this week.

cargo-dylint/tests/ci.rs
```diff
@@ -133,6 +133,165 @@ fn cargo_dylint_and_dylint_readmes_are_equal() {
...
+        .parent()
+        .and_then(|path| path.file_name())
+        .unwrap()
+        .to_string_lossy()
+        .to_string();
+    
+    // Extract the description using regex
+    let re = Regex::new(r#"description\s*=\s*"([^"]*)""#).unwrap();
+    let description = if let Some(caps) = re.captures(&content) {
+        if let Some(desc) = caps.get(1) {
```
smoelius, line 253:
```suggestion
        let desc = caps.get(1).unwrap();
```
If the `captures` call succeeds, then this call should succeed as well. 

A failure of this `get` would likely indicate a bug in the definition of `re`, which I think is worth a test panic.

Note that removing this `if` will require other changes.

cargo-dylint/tests/ci.rs
```diff
@@ -133,6 +133,165 @@ fn cargo_dylint_and_dylint_readmes_are_equal() {
...
+            }
+        }
+    }
+    
+    // Sort examples by name
+    examples.sort_by(|(a_0, _), (b_0, _)| a_0.cmp(b_0));
+    examples
+}
+
+fn extract_name_and_description(cargo_path: &Path) -> Option<(String, String)> {
```
smoelius, line 239:
```suggestion
fn extract_name_and_description(cargo_toml_path: &Path) -> Option<(String, String)> {
```
`cargo_path` sounds like it could refer to the `cargo` binary.

cargo-dylint/tests/ci.rs
```diff
@@ -133,6 +133,165 @@ fn cargo_dylint_and_dylint_readmes_are_equal() {
...
+            if cargo_path.exists() {
+                if let Some((name, desc)) = extract_name_and_description(&cargo_path) {
+                    examples.push((name, desc));
+                }
+            }
+        }
+    }
+    
+    // Sort examples by name
+    examples.sort_by(|(a_0, _), (b_0, _)| a_0.cmp(b_0));
```
smoelius, line 235:
```suggestion
    examples.sort_by(|(a, _), (b, _)| a.cmp(b));
```

cargo-dylint/tests/ci.rs
```diff
@@ -133,6 +133,165 @@ fn cargo_dylint_and_dylint_readmes_are_equal() {
...
+        let examples = collect_examples_from_category(examples_dir, category);
+        for (name, description) in examples {
+            writeln!(content, "| [`{name}`](./{category}/{name}) | {description} |").unwrap();
+        }
+    }
+    
+    content
+}
+
+fn extract_between_markers(content: &str, start_marker: &str, end_marker: &str) -> Option<String> {
```
smoelius, line 211:
```suggestion
fn extract_between_markers(content: &str) -> Option<String> {
    const START_MARKER: &str = "<!-- lint descriptions start -->";
    const END_MARKER: &str = "<!-- lint descriptions end -->";
```
Since `extract_between_markers` is called only once, could we make `start_marker` and `end_marker` constants defined within this function?

cargo-dylint/tests/ci.rs
```diff
@@ -133,6 +133,165 @@ fn cargo_dylint_and_dylint_readmes_are_equal() {
...
+        "experimental",
+        "testing",
+    ];
+    
+    // Read the existing README
+    let readme_path = examples_dir.join("README.md");
+    let readme_content = read_to_string(&readme_path).unwrap();
+
+    // Generate just the lint description tables
+    let tables = generate_lint_tables(&examples_dir, &categories);
```
smoelius, line 152:
```suggestion
    let expected_tables = generate_lint_tables(&examples_dir, &categories);
```
Nit.

cargo-dylint/tests/ci.rs
```diff
@@ -133,6 +133,165 @@ fn cargo_dylint_and_dylint_readmes_are_equal() {
...
+        "<!-- lint descriptions end -->"
+    ).unwrap_or_else(|| panic!("Lint description markers not found in README.md"));
+
+    // Extract and compare just the meaningful content (links and descriptions)
+    let expected_content = extract_meaningful_content(&tables);
+    let actual_content = extract_meaningful_content(&current_tables);
+
+    assert_eq!(
+        expected_content,
+        actual_content,
```
smoelius, line 166:
```suggestion
        expected_tables,
        actual_tables,
```
I would prefer this more stringent check. Please also remove the `extract_meaningful_content` function.

I understand why you wrote the `extract_meaningful_content` function: to avoid having to worry about spacing.

But generating a table with correct spacing will make it easier to replace an existing, incorrect table.

cargo-dylint/tests/ci.rs
```diff
@@ -133,6 +133,165 @@ fn cargo_dylint_and_dylint_readmes_are_equal() {
...
+    
+    // Read the existing README
+    let readme_path = examples_dir.join("README.md");
+    let readme_content = read_to_string(&readme_path).unwrap();
+
+    // Generate just the lint description tables
+    let tables = generate_lint_tables(&examples_dir, &categories);
+    
+    // Extract the current tables section from README using markers
+    let current_tables = extract_between_markers(&readme_content,
```
smoelius, line 155:
```suggestion
    let actual_tables = extract_between_markers(&readme_content, 
```
Nit to follow the convention that "expected" is often paired with "actual".

cargo-dylint/tests/ci.rs
```diff
@@ -133,6 +133,165 @@ fn cargo_dylint_and_dylint_readmes_are_equal() {
...
+                desc_cap.get(1).unwrap().as_str().trim().to_string()
+            ));
+        }
+    }
+    
+    result.sort();
+    result
+}
+
+fn generate_lint_tables(examples_dir: &Path, categories: &[&str]) -> String {
```
smoelius, line 191: As per the comment above regarding `assert_eq!`, this function will need to be adjusted to get the spacing right.

I recommend computing all of the formatted "Example" and "Description/check" entries, and then determining the columns' widths from that.

Something similar to this: https://github.com/trailofbits/dylint/blob/67eddafd13de22a77b3a90922f74f06742cb8174/dylint/src/lib.rs#L182-L193

smoelius (commented):
@jackcooper20 This is looking pretty good. Thanks a lot for working on this!

jackcooper20:
@smoelius I've fixed the things you mentioned in comments, but looks like the CI is failing again for the reason I mentioned previously. I am writing it here again. Please have a look

There might be a potential bug in the `update_example_READMEs.sh` script.

In particular, the problem seems to be with [this line](https://github.com/trailofbits/dylint/blob/master/scripts/update_example_READMEs.sh#L34)
```bash
sed 's/| A lint to check for \([^|]*\) |$/| \u\1 |/'
```

The intent of this line is to capitalize text (e.g., changing `abc` to `Abc`). However, the `\u` syntax isn't supported in GNU `sed`, which results in an extra `u` being introduced before the text (e.g., `uLint description` instead of `Lint description`). 

This issue is likely causing a mismatch between the expected and actual content in `README.md`.

jackcooper20:
Hey @smoelius, I've managed to pass the CI. Apparently there were some new lines that was causing the problem. Please check and let me know if you have any questions.

Thank You!

jackcooper20:
Hello @smoelius, thanks for the feedback.

Apologies for making a mess of it. To be honest, after all the CI failures, I started getting a bit nervous and ended up trying all kinds of approaches just to get things working. That led to a few missteps, and I agree the recent changes added more confusion than value.

I'll revert back to `3589ae5` and work from there with a more simplified approach.

Thanks again for your patience — I'll follow up soon with a cleaner version of the PR.

cargo-dylint/tests/ci.rs
```diff
@@ -133,6 +133,169 @@ fn cargo_dylint_and_dylint_readmes_are_equal() {
...
+    for category in categories {
+        use std::cmp::max;
+        use std::fmt::Write;
+
+        // Get the examples for this category
+        let examples = collect_examples_from_category(examples_dir, category);
+
+        // Calculate column widths
+        let example_header = "Example";
+        let desc_header = "Description/check";
```
smoelius, line 179:
```suggestion
        const EXAMPLE_HEADER: &str = "Example";
        const DESC_HEADER: &str = "Description/check";
```
Nit. And these will probably have to be moved to earlier in the function body.

cargo-dylint/tests/ci.rs
```diff
@@ -133,6 +133,169 @@ fn cargo_dylint_and_dylint_readmes_are_equal() {
...
+    // Get the name from the directory
+    let name = cargo_toml_path
+        .parent()
+        .and_then(|path| path.file_name())
+        .unwrap()
+        .to_string_lossy()
+        .to_string();
+
+    // Extract the description using regex
+    let re = Regex::new(r#"description\s*=\s*"([^"]*)""#).unwrap();
```
smoelius, line 270: Since `extract_name_and_description` will be called multiple times, could you please make this a static using `LazyLock`?

Here are a couple of examples: https://github.com/trailofbits/dylint/blob/d2876b4fb978a13daf0c0c7c24bc307207fcb82e/examples/supplementary/commented_code/src/lib.rs#L102-L105

smoelius (commented):
I think this looks pretty good. I have just a couple of small nits below.

Also, just for my sanity, could you please rebase this onto master so that we verify it passes there?

jackcooper20:
@smoelius implemented they changes you requested. Please have a look. Thank You!

smoelius:
Looks great. Could you please squash your commits down to one? Then I will merge this.

jackcooper20:
@smoelius, squashed all the commits.

smoelius:
Thanks, @jackcooper20!

Issue #1358, opened 2024-10-03
`abs_home_path`: don't warn in build or test contexts


smoelius:
Such warnings tend to be false positives.

Examples where the lint is allowed:
- [x] Build context: https://github.com/trailofbits/dylint/blob/4438f402591d05e369636f54e18b4bab232702b5/dylint/build.rs#L23
- [ ] Test context: https://github.com/trailofbits/dylint/blob/4438f402591d05e369636f54e18b4bab232702b5/cargo-dylint/tests/integration/fix.rs#L82

smoelius:
Clippy uses the crate name to detect build scripts: https://github.com/rust-lang/rust-clippy/blob/4f0e46b74dbc8441daf084b6f141a7fe414672a2/clippy_lints/src/write.rs#L293-L298

nowak-token:
Hey @smoelius as the last issue on necessist requires further comment, in the mean time is this issue still available? if yes then maybe I can open a pr here?

smoelius:
Would you want to tackle just one of the two cases to start?

nowak-token:
@smoelius sure, let's start with one case, if its successful then I'll start working on the other.
How about I start with the build one first? what do you think?

smoelius:
Sounds good to me. Thank you!

nowak-token:
@smoelius I've tried to solve the first part of the issue in PR #1569 , please take a look at it, and advise me if I'm on the right path. Thanks.

nowak-token:
@smoelius are you working on the test context of this issue? If its free then I'd like to send a  pr to resolve this too.

smoelius:
@nowak-token You are welcome to work on this.

I would like to suggest using [`clippy_utils::is_in_test`](https://doc.rust-lang.org/stable/nightly-rustc/clippy_utils/fn.is_in_test.html) as part of the solution.

Also, if it's not a burden, please include a test with the fix. (As you probably recall, I forgot to request a test up front in #1569.)

PR #1569, opened 2025-04-03
Don't warn about absolute home paths in build scripts

nowak-token:
## Description
This PR implements the first part of issue #1358, making the `abs_home_path` lint smarter by not warning in build script contexts. Build scripts often need to reference absolute paths during the build process, making these warnings false positives.

## Changes
- Added `rustc_session` to the extern crate declarations
- Added detection of build scripts in the `check_expr` method by checking if the crate name is "build_script_build"
- Skipped lint checking when in a build script context

## Implementation details
The implementation follows Clippy's approach of checking the crate name to detect build scripts, as referenced in the issue.

## Testing
I've verified the lint no longer triggers in build scripts by testing against the examples in the issue.

## Related Issue
Resolves first part of #1358

### DIFF
--- a/examples/general/abs_home_path/src/lib.rs
+++ b/examples/general/abs_home_path/src/lib.rs
@@ -11,7 +11,7 @@ use dylint_internal::{home, paths};
 use once_cell::unsync::OnceCell;
 use rustc_ast::ast::LitKind;
 use rustc_hir::{Closure, Expr, ExprKind, Item, ItemKind, Node, def_id::DefId};
-use rustc_lint::{LateContext, LateLintPass};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_span::Span;
 use std::{
     collections::HashSet,
@@ -31,6 +31,10 @@ dylint_linting::impl_late_lint! {
     /// ### Known problems
     ///
     /// The lint does not apply inside macro arguments. So false negatives could result.
+    ///
+    /// ### Note
+    ///
+    /// This lint doesn't warn in build scripts (`build.rs`), as they often need to reference absolute paths.
     ///
     /// ### Example
     ///
@@ -67,6 +71,17 @@ impl<'tcx> LateLintPass<'tcx> for AbsHomePath {
     }
 
     fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr) {
+        // Skip build scripts
+        if cx
+            .sess()
+            .opts
+            .crate_name
+            .as_ref()
+            .is_some_and(|crate_name| crate_name == "build_script_build")
+        {
+            return;
+        }
+
         if cx
             .tcx
             .hir()
--- a/examples/general/abs_home_path/README.md
+++ b/examples/general/abs_home_path/README.md
@@ -13,6 +13,10 @@ The path might not exist when the code is used in production.
 
 The lint does not apply inside macro arguments. So false negatives could result.
 
+### Note
+
+This lint doesn't warn in build scripts (`build.rs`), as they often need to reference absolute paths.
+
 ### Example
 
 ```rust


nowak-token:
@smoelius I'll try to get all the ci tests to pass, regarding that do you have any suggestions you'd like to share?

smoelius:
> @smoelius I'll try to get all the ci tests to pass, regarding that do you have any suggestions you'd like to share?

Just that the `architectures_are_current` test failure is not your PR's fault. I will push a fix for that later on today.

examples/general/abs_home_path/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 
 extern crate rustc_ast;
 extern crate rustc_hir;
+extern crate rustc_session;
```
smoelius, line 7:
```suggestion
```

This should be declared by the `dylint_linting::impl_late_lint!` macro.

examples/general/abs_home_path/src/lib.rs
```diff
@@ -67,6 +68,18 @@ impl<'tcx> LateLintPass<'tcx> for AbsHomePath {
     }
 
     fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr) {
+        // Skip in build scripts
```
smoelius, line 71:
```suggestion
        // Skip build scripts
```

examples/general/abs_home_path/src/lib.rs
```diff
@@ -67,6 +68,18 @@ impl<'tcx> LateLintPass<'tcx> for AbsHomePath {
...
+        let is_build_script = cx
+            .sess()
+            .opts
+            .crate_name
+            .as_ref()
+            .map_or(false, |crate_name| crate_name == "build_script_build");
+            
+        if is_build_script {
+            return;
+        }
```
smoelius, line 81: Could the `let` and this `if` be combined so that there is no declaration and just `if cx.sess()...`?

smoelius (commented):
Would you be willing to add a test for this too? You can say "no".

examples/general/abs_home_path/src/lib.rs
nowak-token: Thanks for the suggestion @smoelius , I had already removed it, but got another error while running cargo test locally `failures:
    test::multiple_libraries_one_toolchain`
    I'm trying to fix this right now, if its successful, I'll commit the changes

nowak-token:
hopefully this fixes all the errors I got in the CI

nowak-token:
@smoelius I've managed to get the CI to pass, please have a look and see if this works to solve the build part of the issue?

smoelius:
> @smoelius I've managed to get the CI to pass, please have a look and see if this works to solve the build part of the issue?

I'll respond shortly.

examples/general/abs_home_path/src/lib.rs
```diff
@@ -6,7 +6,7 @@ extern crate rustc_ast;
 extern crate rustc_hir;
 extern crate rustc_span;
 
-use clippy_utils::{diagnostics::span_lint, match_def_path, path_def_id};
+use clippy_utils::{diagnostics::span_lint, match_def_path, path_def_id, source::HasSession};
```
smoelius, line 9: This is for the `sess` method?

Could we please use [`LintContext`](https://doc.rust-lang.org/beta/nightly-rustc/rustc_lint/trait.LintContext.html) instead?

I think `LintContext` is more widely used and is less likely to change.

examples/general/abs_home_path/src/lib.rs
```diff
@@ -162,3 +177,19 @@ fn ui() {
         .rustc_flags(["--test"])
         .run();
 }
+
+// Test to verify that lint isn't triggered in build scripts
+#[test]
+#[allow(clippy::assertions_on_constants)]
+fn build_script_detection_works() {
```
smoelius, line 184: We will need a test eventually, but I don't think this tests what we want.

The test we want would have a build script with `env!("CARGO_MANIFEST_PATH")` or something like that.

Actually, here's an example in Dylint: https://github.com/trailofbits/dylint/blob/5aad720e3cb54c93dfa2ad7c0d5e7a0cb29c3cf4/dylint/build.rs#L23-L24

I put you on the spot to add a test, which was unfair.

Could I suggest removing this test and we just merge with no test?

smoelius (commented):
If you agree with all of my suggestions, could you please squash your commits down to one?

nowak-token:
@smoelius please check now, everything should be resolved.

nowak-token:
@smoelius can this be merged now? or do i need to make any more changes?

PR #1572, opened 2025-04-05
Add test for abs_home_path lint's build script allowance

### LINKED ISSUE #1571: Need a test for `abs_home_path` allowance in build scripts
@nowak-token very kindly fixed #1358 for build scripts. However, the fix needs a test.

The test should involve a package with a build script. The build script should refer to a path in `$HOME`, say with `env!(CARGO_MANIFEST_DIR)`.

The package doesn't have to do anything meaningful.

For examples of tests that use a package, the `crate_wide_allow` library has several: https://github.com/trailofbits/dylint/blob/f777afac4ea26346de99a6c4f6265cf990d5e926/examples/general/crate_wide_allow/src/lib.rs#L180-L219

This is the package that it uses: https://github.com/trailofbits/dylint/tree/f777afac4ea26346de99a6c4f6265cf990d5e926/examples/general/crate_wide_allow/ui_manifest

tomaswoj-eth:
## Description
This PR adds a test for the abs_home_path lint's allowance in build scripts, addressing issue #1571. The implementation of this allowance was done in another PR but a test was needed to verify the fix works correctly.

## Changes
- Created a test package `ui_build_script` with a build script that references an absolute path
- Added a test function `build_script_allowance()` that verifies the lint doesn't trigger for build scripts
- Added documentation explaining the purpose of the test

## Test Coverage
The test verifies that when the abs_home_path lint runs on a package with a build script containing absolute paths, the lint correctly ignores those paths as intended.

## Fixes
Fixes #1571


tomaswoj-eth:
Hey @smoelius I'm have trouble getting this particular [CI / test (ubuntu-latest, cargo-dylint-ci) (pull_request)](https://github.com/trailofbits/dylint/actions/runs/14278033108/job/40024029720?pr=1572) to pass could you please take a look at it and give me some advice on how I can fix this?

smoelius:
I'm glad to see that you got it working. I'll review this within the next few days.

My initial impressions are: this looks pretty good!

examples/general/abs_home_path/src/lib.rs
```diff
@@ -177,3 +177,42 @@ fn ui() {
...
+        return;
+    }
+
+    // Run cargo check with our lint on the test package
+    let output = Command::new("cargo")
+        .args(["check", "--manifest-path", "ui_build_script/Cargo.toml"])
+        .env("DYLINT_LIBRARY_PATH", env!("CARGO_MANIFEST_DIR"))
+        .current_dir(env!("CARGO_MANIFEST_DIR"))
+        .logged_output(true)
+        .expect("Failed to execute cargo command");
```
smoelius, line 209: I don't think this runs the lint on the test package.

If you look at this other `crate_wide_allow` test, it builds `cargo-dylint` and then runs that: https://github.com/trailofbits/dylint/blob/f777afac4ea26346de99a6c4f6265cf990d5e926/examples/general/crate_wide_allow/src/lib.rs#L140-L172

I think we need to do something similar here.

A good sanity check is to comment out the fix that was introduced by #1569, and see if the lint triggers: https://github.com/trailofbits/dylint/blob/67eddafd13de22a77b3a90922f74f06742cb8174/examples/general/abs_home_path/src/lib.rs#L75-L83

Currently, if you comment out that code and run the test, the lint does not trigger.

Also, please add the following to the `Command` that will run `cargo-dylint`:
```rs
.env("DYLINT_RUSTFLAGS", "--deny warnings")
```
This will cause `cargo-dylint` to exit with an error if any lint triggers. In this way, `assert!(output.status.success(), ...)` ensures that no lints triggered.

examples/general/abs_home_path/ui_build_script/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "ui_build_script"
+version = "4.1.0"
+edition = "2021"
+build = "build.rs"
```
smoelius, line 5:
```suggestion
```
I don't think this is necessary. `build.rs` should be automatically identified as a build script because of its file name.

smoelius (commented):
This test has turned out to be trickier than I imagined.

Your test package definitely has the right structure. I just don't we're invoking Dylint on it in the right way now.

Thanks a lot for working on this.

smoelius:
@tomaswoj-eth Would you mind putting this PR on hold for hopefully no more than a week?

Getting access to the current `cargo-dylint` should be easy. Hence, the repo should provide an internal function that does that.

Would you mind pausing this PR while I add such an internal function?

tomaswoj-eth:
@smoelius sure I'll wait on this while you add the necessary functions, meanwhile is it possible to work on the second part of #1358 ? or will that also be affected by the function you will add and need to wait?

smoelius:
@tomaswoj-eth Thanks very much for your patience.

In #1580, I added a function a function `dylint_internal::testing::cargo_dylint` to debug-build `cargo-dylint` and return a path to the executable: https://github.com/trailofbits/dylint/blob/e37e757b95e9e16fe4f214e7bdac78b272ddca43/internal/src/testing.rs#L28

I think you should be able to change this code: https://github.com/trailofbits/dylint/blob/a48226bd31d5abdb5623f7e4cbb8b085c639bcbb/examples/general/abs_home_path/src/lib.rs#L204-L207
to something like this:
```rs
    let cargo_dylint = dylint_internal::testing::cargo_dylint("../../..").unwrap();
    let output = Command::new(cargo_dylint)
        .env("DYLINT_RUSTFLAGS", "--deny warnings")
        .args([
            "dylint",
            "--manifest-path",
            "ui_build_script/Cargo.toml",
            "--path",
            env!("CARGO_MANIFEST_DIR")
        ]) 
```
and it should "just work". 🤞

Just to recount what I said in https://github.com/trailofbits/dylint/pull/1572#discussion_r2032069796, the command should:
- succeed with the code as is
- fail if the fix from #1569 is commented out

tomaswoj-eth:
@smoelius I've implemented the changes you suggested, using `dylint_internal::testing::cargo_dylint` and updating the test accordingly. The test passes locally after I added the "testing" feature to the `dylint_internal` dependency.

However, I notice the CI is failing with an error in the `dylint-link` package:
```rs
test::architectures_are_current ... FAILED
```

The error message mentions workspace configuration issues with a temporary directory:
```rs
warning: compiling this new package may not work due to invalid workspace configuration
current package believes it's in a workspace when it's not:
current: /home/runner/work/dylint/dylint/dylint-link/.tmpEtuaiA/Cargo.toml
workspace: /home/runner/work/dylint/dylint/Cargo.toml
```

Since this is failing in a different component than the one I modified, and it's related to workspace configuration rather than lint behavior, I believe this might be unrelated to my changes. Could you advise on how to proceed with this CI error?

smoelius:
I agree that failure is unlikely to be related to your changes, but I also don't know what's going on there.

I will try to look into it later today.

Thanks again for your patience.

smoelius:
I think the CI failure was some sort of cache weirdness.

I'll give this PR another look within the next few days.

smoelius (commented):
I think we're just about there.

I you could, please:
- Address the outstanding GitHub comments.
- Run `cargo fmt` so that the recently added `fmt` test is happy.
- Remove the `smoelius: ` prefixes from the comments that have them. (I use that to prefix comments I write, so that people know I wrote them.)
- Commit the lockfile that this test generates.
- Squash all of your commits down to one.

tomaswoj-eth:
@smoelius please check now

examples/general/abs_home_path/ui_build_script/Cargo.toml
smoelius: Could you please apply this change? Then, I think this should be good to go.

tomaswoj-eth: @smoelius done. Sorry I missed this before.

tomaswoj-eth:
@smoelius I think this can be cleanly merged now.

smoelius:
Beautiful! Thank you very much for doing this!

PR #1579, opened 2025-04-08
Add: format test

### LINKED ISSUE #1578: Need a `format` test
> Following #1529, the "Format" task was incorrectly marked as completed.
> 
> "Incorrectly" because `rustdoc-prettier` does not subsume `cargo fmt`. Hence, something like the following from format.sh (now removed) must be re-added in the form of a Rust test: https://github.com/trailofbits/dylint/blob/0fab20fb1757079eb0bf7fd6bb8901ea1e949ea5/scripts/format.sh#L32-L33
> 
> The following could serve as a basis: https://github.com/trailofbits/cargo-unmaintained/blob/0d4e3b96167cf5edfaded0e3a09033a79debe6e4/ci/src/lib.rs#L51-L57
> 
> However, it should be adapted to use `walkdir`, much like `sort` does: https://github.com/trailofbits/dylint/blob/67eddafd13de22a77b3a90922f74f06742cb8174/cargo-dylint/tests/ci.rs#L499-L510
> 
> (Actually, this should be pretty simple.) 

 _Originally posted by @smoelius in [#636](https://github.com/trailofbits/dylint/issues/636#issuecomment-2784663328)_

YanVictorSN:
Hey @smoelius. I have two questions about this PR:

Should I use `--check` like the sort function or do we need to modify the files directly?

Also, we need to use `preserves_cleanliness()` before to follow the pattern used in the rest of the file?

Closes #1578

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,10 +75,6 @@ jobs:
         run: |
           # smoelius: Prettier is still needed for scripts/update_example_READMEs.sh.
           npm install -g prettier
-          rm -f "$HOME"/.cargo/bin/cargo-fmt
-          rm -f "$HOME"/.cargo/bin/rustfmt
-          rustup install nightly
-          rustup component add rustfmt --toolchain nightly
 
       - name: Format example READMEs
         run: ./scripts/update_example_READMEs.sh && git diff --exit-code
@@ -142,6 +138,7 @@ jobs:
         run: |
           npm install -g prettier
           rustup install nightly
+          rustup component add rustfmt --toolchain nightly
 
       # smoelius: This list will grow: https://github.com/trailofbits/dylint/issues/636
       - name: Install CI tools
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -527,9 +527,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.32"
+version = "4.5.34"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6088f3ae8c3608d19260cd7445411865a485688711b78b5be70d78cd96136f83"
+checksum = "e958897981290da2a852763fe9cdb89cd36977a5d729023127095fa94d95e2ff"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -537,9 +537,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.5.32"
+version = "4.5.34"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "22a7ef7f676155edfb82daa97f99441f3ebf4a58d5e32f295a56259f1b6facc8"
+checksum = "83b0f35019843db2160b5bb19ae09b4e6411ac33fc6a712003c33e03090e2489"
 dependencies = [
  "anstream",
  "anstyle",
@@ -2231,9 +2231,9 @@ dependencies = [
 
 [[package]]
 name = "icu_locid_transform_data"
-version = "1.5.0"
+version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fdc8ff3388f852bede6b579ad4e978ab004f139284d7b28715f773507b946f6e"
+checksum = "7515e6d781098bf9f7205ab3fc7e9709d34554ae0b21ddbcb5febfa4bc7df11d"
 
 [[package]]
 name = "icu_normalizer"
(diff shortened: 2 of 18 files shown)


smoelius:
@YanVictorSN Thanks for jumping on this so quickly!

To answer your questions:

> Should I use `--check` like the sort function or do we need to modify the files directly?

Please use `--check`.

> Also, we need to use `preserves_cleanliness()` before to follow the pattern used in the rest of the file?

Since `--check` will be used, `preserves_cleanliness()` should not be necessary.

Thanks again!

YanVictorSN:
@smoelius Always very receptive, thank you!

I just made the changes, but I have some new doubts.

Do I need to add `cargo-fmt` to the CI workflow? I noticed that in the `workflow/ci.yml` file we already have this line:

`rustup component add rustfmt --toolchain nightly
`

But I'm still getting this error in the GitHub CLI:

`error: 'cargo-fmt' is not installed for the toolchain 'nightly-x86_64-unknown-linux-gnu`

So I'm a bit confused. Can you help me?

smoelius:
I'm not sure what's going on, but I'll look into it.

YanVictorSN:
> I'm not sure what's going on, but I'll look into it.

Thanks!

smoelius:
Do you mind if I push to your branch?

When I tested this locally, I experienced other failures that I had to fix. (Those failures were faults of my issue description, not of your PR.)

The main change I want to make is to change the directories in which `cargo fmt` is run rather than use `--manifest-path`.

The changes are minor, but I think it would be easier to just push them than to explain each change.

So is that okay?

smoelius:
I'm going to have to dig into this more.

Sorry, @YanVictorSN. I thought this was going to be an easy addition.

YanVictorSN:
> I'm going to have to dig into this more.
> 
> Sorry, @YanVictorSN. I thought this was going to be an easy addition.

That's fine @smoelius ! Looks like the CI passed. I noticed that you removed the check to skip the root directory that you mentioned in the command: `-mindepth 2` and put `!dbg(parent)` instead. So now we're checking all the `Cargo.toml` files, is that ok?

smoelius:
> That's fine @smoelius ! Looks like the CI passed.

`rustfmt` was being added for the nightly toolchain in the [`lint`](https://github.com/trailofbits/dylint/blob/06cf5bdd32ecdc2001c077a87afe784b3b2d96d6/.github/workflows/ci.yml#L50) job. But this PR moves the need for that component from the  `lint` job to the [`test`](https://github.com/trailofbits/dylint/blob/06cf5bdd32ecdc2001c077a87afe784b3b2d96d6/.github/workflows/ci.yml#L89) job, where the component was not being added. I should have spotted this problem much sooner.

> I noticed that you removed the check to skip the root directory that you mentioned in the command: `-mindepth 2` and put `!dbg(parent)` instead.

My mistake, I should have removed that `dbg!`.

> So now we're checking all the `Cargo.toml` files, is that ok?

We're actually checking all of the directories containing Cargo.toml files, and that appears to be okay.

I'm not 100% sure, but I think `cargo fmt --manifest-path PATH` requires `PATH` to not refer to a virtual manifest, and that was what was causing problems before.

Before we merge this, can I ask why you included this `windows` prohibition? https://github.com/trailofbits/dylint/blob/f7dfc1690617f9959b5366ff9bbd52485831f61b/cargo-dylint/tests/ci.rs#L486

YanVictorSN:
> `rustfmt` was being added for the nightly toolchain in the [`lint`](https://github.com/trailofbits/dylint/blob/06cf5bdd32ecdc2001c077a87afe784b3b2d96d6/.github/workflows/ci.yml#L50) job. But this PR moves the need for that component from the `lint` job to the [`test`](https://github.com/trailofbits/dylint/blob/06cf5bdd32ecdc2001c077a87afe784b3b2d96d6/.github/workflows/ci.yml#L89) job, where the component was not being added. I should have spotted this problem much sooner.

> We're actually checking all of the directories containing Cargo.toml files, and that appears to be okay.
> 
> I'm not 100% sure, but I think `cargo fmt --manifest-path PATH` requires `PATH` to not refer to a virtual manifest, and that was what was causing problems before.
> 

Thanks for the explanation!

>Before we merge this, can I ask why you included this windows prohibition?
>https://github.com/trailofbits/dylint/blob/f7dfc1690617f9959b5366ff9bbd52485831f61b/cargo-dylint/tests/ci.rs#L486

My mistake. I should have looked more carefully before putting this in. 

I was copying the pattern from the prettier functions, but I see now that we use that because of the backslash in the args, which can cause issues on Windows.

I think we don’t need it in `fmt`.

smoelius:
> I think we don’t need it in `fmt`.

Would you mind removing it, and then I will merge?

YanVictorSN:
> > I think we don’t need it in `fmt`.
> 
> Would you mind removing it, and then I will merge?

Not at all. Just did, thanks!

PR #1603, opened 2025-04-22
Fix: Don't trigger lint for std::fmt::Result return types

### LINKED ISSUE #929: `non_local_effect_before_error_return`: don't fire in functions returning `std::fmt::Result`
The reports tend to be uninteresting.

nowak-token:
## Description

This PR addresses issue #929 by preventing the `non_local_effect_before_error_return` lint from firing in functions that return `std::fmt::Result`.

### Problem

The lint was previously triggering warnings for functions returning `std::fmt::Result`, but as noted in the issue, these reports tend to be uninteresting and add unnecessary noise to the lint output.

### Changes

1. Renamed the `is_result` function to `is_lintable_result` to better reflect its purpose
2. Enhanced the function to check if the second generic argument of `Result` is `core::fmt::Error`
3. If the error type is `core::fmt::Error`, the function now returns `false`, preventing the lint from being applied
4. Added test cases in both `ui/main.rs` and `ui_public_only/main.rs` to verify the behavior

### Implementation Details

The implementation follows the approach suggested in the issue:
- Access the generic arguments (`substs`) in the Result type
- Check the second argument (index 1) using `clippy_utils::match_def_path` to determine if it's `core::fmt::Error`
- Return `false` in this case to skip applying the lint

### Testing

Added test functions returning `fmt::Result` with non-local effects before returning an error to verify the fix. The tests pass with no warnings being emitted for these functions, confirming the lint no longer fires for `std::fmt::Result` return types.

Fixes #929

### DIFF
--- a/examples/general/non_local_effect_before_error_return/src/lib.rs
+++ b/examples/general/non_local_effect_before_error_return/src/lib.rs
@@ -150,7 +150,7 @@ impl<'tcx> LateLintPass<'tcx> for NonLocalEffectBeforeErrorReturn {
             return;
         }
 
-        if !is_result(cx, cx.typeck_results().expr_ty(body.value)) {
+        if !is_lintable_result(cx, cx.typeck_results().expr_ty(body.value)) {
             return;
         }
 
@@ -221,9 +221,21 @@ fn in_async_function(tcx: ty::TyCtxt<'_>, hir_id: rustc_hir::HirId) -> bool {
         })
 }
 
-fn is_result(cx: &LateContext<'_>, ty: ty::Ty) -> bool {
-    if let ty::Adt(adt, _) = ty.kind() {
-        cx.tcx.is_diagnostic_item(sym::Result, adt.did())
+fn is_lintable_result(cx: &LateContext<'_>, ty: ty::Ty) -> bool {
+    if let ty::Adt(adt, substs) = ty.kind() {
+        if !cx.tcx.is_diagnostic_item(sym::Result, adt.did()) {
+            return false;
+        }
+
+        // Don't lint if the error type is core::fmt::Error
+        if let Some(error_ty) = substs.get(1)
+            && let ty::Adt(error_adt, _) = error_ty.expect_ty().kind()
+            && match_def_path(cx, error_adt.did(), &["core", "fmt", "Error"])
+        {
+            return false;
+        }
+
+        true
     } else {
         false
     }
--- a/examples/general/non_local_effect_before_error_return/ui/main.rs
+++ b/examples/general/non_local_effect_before_error_return/ui/main.rs
@@ -276,3 +276,15 @@ pub mod public_only {
         Err(VarError::NotPresent)
     }
 }
+
+// Test to check that functions returning std::fmt::Result should not trigger the lint
+pub mod fmt_result_test {
+    use std::fmt::{self, Write};
+
+    pub fn fmt_result_with_write_before_err_return(buffer: &mut String) -> fmt::Result {
+        // This non-local effect (write) should not trigger a lint warning
+        // because the return type is std::fmt::Result
+        buffer.write_str("Hello, world!")?;
+        Err(fmt::Error)
+    }
+}
--- a/examples/general/non_local_effect_before_error_return/ui/main.stderr
+++ b/examples/general/non_local_effect_before_error_return/ui/main.stderr
@@ -155,12 +155,6 @@ warning: assignment to dereference before error return
 LL |         *flag = true;
    |         ^^^^^^^^^^^^
 
-warning: call to `std::fmt::DebugStruct::<'_, '_>::field` with mutable reference before error return
-  --> $DIR/main.rs:260:12
-   |
-LL | pub struct Foo {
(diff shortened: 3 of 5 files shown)


nowak-token:
@smoelius you seem to have a lot on your plate right now. So, check this when you're free, no rush.

nowak-token:
@smoelius let me know when you're free to review this.

examples/general/non_local_effect_before_error_return/src/lib.rs
```diff
@@ -221,9 +221,22 @@ fn in_async_function(tcx: ty::TyCtxt<'_>, hir_id: rustc_hir::HirId) -> bool {
...
+fn is_permitted_result(cx: &LateContext<'_>, ty: ty::Ty) -> bool {
+    if let ty::Adt(adt, substs) = ty.kind() {
+        if !cx.tcx.is_diagnostic_item(sym::Result, adt.did()) {
+            return false;
+        }
+
+        // Don't lint if the error type is std::fmt::Error
+        if let Some(error_ty) = substs.get(1) {
+            if let ty::Adt(error_adt, _) = error_ty.expect_ty().kind() {
+                if match_def_path(cx, error_adt.did(), &["std", "fmt", "Error"]) {
```
smoelius, line 233:
```suggestion
        if let Some(error_ty) = substs.get(1)
            && let ty::Adt(error_adt, _) = error_ty.expect_ty().kind()
            && match_def_path(cx, error_adt.did(), &["std", "fmt", "Error"]) {
```
Nit. Note some of the surrounding code will have to change.

examples/general/non_local_effect_before_error_return/src/lib.rs
```diff
@@ -150,7 +150,7 @@ impl<'tcx> LateLintPass<'tcx> for NonLocalEffectBeforeErrorReturn {
             return;
         }
 
-        if !is_result(cx, cx.typeck_results().expr_ty(body.value)) {
+        if !is_permitted_result(cx, cx.typeck_results().expr_ty(body.value)) {
```
smoelius, line 153:
```suggestion
        if !is_disallowed_result(cx, cx.typeck_results().expr_ty(body.value)) {
```
I know I suggested `is_permitted_result` in #929, but I see now that it doesn't make sense.

We want to continue if type of the body is a result that is _not_ permitted.

So I am suggesting now `is_disallowed result`, but I don't love it.

Do you have a name you think is more apt?

examples/general/non_local_effect_before_error_return/ui/main.rs
```diff
@@ -276,3 +276,15 @@ pub mod public_only {
...
 }
+
+// Test for the fix in issue #929: functions returning std::fmt::Result should not trigger the lint
+pub mod fmt_result_test {
+    use std::fmt::{self, Write};
+
+    pub fn fmt_result_with_write_before_err_return(buffer: &mut String) -> fmt::Result {
+        // This non-local effect (write) should not trigger a lint warning
+        // because the return type is std::fmt::Result
+        write!(buffer, "Hello, world!")?;
```
smoelius, line 287:
```suggestion
        buffer.write_str("Hello, world!")?;
```
I tried disabling the fix in `is_disallowed_result` to verify that the lint was firing on this line, and it was not.

After some debugging, I realized that the lint was getting to the `span_lint_and_then` call in lib.rs, the warning just wasn't being shown.

I suspect the warning was not being shown because `fn_span` was in the result of a macro expansion, though I have not confirmed this.

Regardless, changing the call to `write_str` causes the warning to be shown then the fix in `is_disallowed_result` is disabled.

examples/general/non_local_effect_before_error_return/src/lib.rs
smoelius, line 233:
```suggestion
                if match_def_path(cx, error_adt.did(), &["core", "fmt", "Error"]) {
```
I figured out that `std` should be `core` by disabling the fix `is_disallowed_result` and checking whether the lint was firing.

smoelius (commented):
@nowak-token I would like give this another look after you have had a chance to review my comments. I hope you don't mind.

examples/general/non_local_effect_before_error_return/src/lib.rs
nowak-token: @smoelius I was thinking something along the lines of `is_lintable_result` or `is_lint_target`. I came up  with these due to the fact that it's clear that we're determining whether to apply the lint or not in this fn. What do you think about it?
For now i'm making changes with `is_disallowed_result` just as a placeholder, let me  know your choice, I'll make another commit to update it.

smoelius: I like `is_lintable_result`. Let's please go with that. :)

nowak-token:
@smoelius take another look just to be sure, although I ran the tests and they passed, but just to be on the safe side have a final look.

smoelius:
I ran the same test that I did yesterday, by verifying that the lint fires on the new test when the fix is disabled. So I am pretty confident that the fix is correct.

Thanks a lot for this!

PR #1604, opened 2025-04-25
Don't warn about absolute home paths in test contexts

### LINKED ISSUE #1358: `abs_home_path`: don't warn in build or test contexts

nowak-token:
## Description
This PR addresses the second part of issue #1358 by adding test context detection to the `abs_home_path` lint. The lint will now skip warning about absolute home paths when they appear in test functions, similar to how it already skips them in build scripts.

## Implementation Details
- Used `clippy_utils::is_in_test` as recommended to detect if code is running in a test context
- Maintained the existing test function detection for compatibility
- Added a test package (`ui_test`) with test functions that use absolute home paths
- Added a `test_context_allowance` test function to verify the fix works correctly

## Motivation
Test functions often need to reference absolute paths (like `CARGO_MANIFEST_DIR`) for test data or file operations. These references trigger false positive warnings as they aren't intended for production use.

## Testing
- Created a test package demonstrating the fix works correctly
- Added a test function that verifies the lint doesn't warn in test contexts
- Ensured all existing tests continue to pass

## Related Issues
Completes the work started in PR #1569, which addressed the build script portion of issue #1358.

Fixes #1358


nowak-token:
@smoelius check if the fix and the tests match with what you expected.

smoelius:
I'll try to get to this soon.

nowak-token:
@smoelius any updates?

smoelius:
I performed the same experiment I did in #1603. I disabled the fix to see whether the lint fires, and it doesn't. I think I know why, but I need to dig into it more.

nowak-token:
@smoelius, Thanks for looking into this and for the comment about disabling the fix. I investigated the behavior you described before.

I reproduced this by temporarily disabling the `is_in_test(cx.tcx, expr.hir_id)` check in `check_expr` and then running `cargo-dylint` directly on the `ui_test` package. The output confirmed your finding: even with that specific check disabled, warnings were only emitted for the non-test function (`non_test_function_with_home_path`), not for the code inside the `#[test]` functions.

It appears there's another mechanism at play, likely related to how `rustc` handles the `--test` flag or how the test harness (`compiletest_rs`) sets up the compilation for UI tests. This underlying mechanism seems to prevent lints from firing within `#[test]` functions before the lint pass even gets to fully evaluate those expressions. This would explain why disabling the specific check didn't make the warnings appear in the test functions during your experiment.

Despite this underlying behavior, I think it's still best to keep the explicit `is_in_test(cx.tcx, expr.hir_id)` check in our `check_expr` as  it aligns with the common practice seen in many Clippy lints, where test contexts are often explicitly checked.

The current version of the code, with the `is_in_test` check included and the subsequent simplifications (removing the redundant `check_crate`, `test_fns` tracking, etc.), now correctly handles the test cases (warnings appear in non-test code, no warnings in test code) and passes all tests, including the UI and `test_context_allowance` checks. I believe this addresses the test context part of issue #1358 as intended.

Or, if you have any suggestions or ideas you want to add please let me know.

examples/general/abs_home_path/src/lib.rs
```diff
@@ -227,3 +173,69 @@ fn build_script_allowance() {
         "The abs_home_path lint was incorrectly triggered in a build script: {stderr}"
     );
 }
+
+#[test]
+fn test_context_allowance() {
```
smoelius, line 178: This test is very similar to `build_script_allowance`.

Could you combine the two so that, in the end, there is one test with a two-iteration loop?

examples/general/abs_home_path/ui_test/src/main.rs
```diff
@@ -0,0 +1,28 @@
...
+    let _ = option_env!("CARGO");
+}
+
+// Another test function to ensure multiple test functions work correctly
+#[test]
+fn another_test_function() {
+    // This should also NOT trigger the lint
+    let home_path = env!("CARGO_MANIFEST_DIR");
+    println!("Home path: {}", home_path);
+}
```
smoelius, line 28: I think it is sufficient to have just one test and to remove this one.

examples/general/abs_home_path/ui_test/src/main.rs
```diff
@@ -0,0 +1,28 @@
+fn main() {
+    println!("Testing abs_home_path lint's test context allowance");
+
+    // This SHOULD trigger the abs_home_path lint because we're NOT in a test function
+    non_test_function_with_home_path();
+}
+
+// This function should trigger the lint because it's not a test function
+fn non_test_function_with_home_path() {
+    // This should trigger the lint when fix is disabled
```
smoelius, line 10: I think it is sufficient to have just one comment here, rather than three.

Maybe just the last one, so that there is symmetry with `test_function_with_home_path`?

smoelius (commented):
> Despite this underlying behavior, I think it's still best to keep the explicit `is_in_test(cx.tcx, expr.hir_id)` check in our `check_expr` as it aligns with the common practice seen in many Clippy lints, where test contexts are often explicitly checked.

I am inclined to agree, and ripping out `find_test_fns`, etc. like you did in a4fd2f9e1f56b29e3e6fd7f08daa2a0dd57ddceb is what I would have suggested.

I made a few comments below. I may have to look at this again. I hope you don't mind.

examples/general/abs_home_path/src/lib.rs
smoelius: Something I forgot to mention: the data structure that you loop over could include a closure to perform the checks after `cargo-dylint` is run.

In this way, the tests could perform different checks.

nowak-token:
@smoelius  are you making a fix for the failing [CI / test (ubuntu-latest, cargo-dylint-ci) (pull_request)](https://github.com/trailofbits/dylint/actions/runs/14728128640/job/41335657523?pr=1604) test? i'll update when the fix is in place.

smoelius:
> @smoelius are you making a fix for the failing [CI / test (ubuntu-latest, cargo-dylint-ci) (pull_request)](https://github.com/trailofbits/dylint/actions/runs/14728128640/job/41335657523?pr=1604) test? i'll update when the fix is in place.

It looks sporadic. I just restarted the test.

nowak-token:
@smoelius I ran the test locally `cargo test --test ci markdown_link_check` and the same error occured there. It's occuring in this run too. I think there is some other issue here. I'm researching this.

nowak-token:
It appears the test is hitting GitHub's rate limits when checking all the links (especially commit links) in `CHANGELOG.md`, resulting in numerous HTTP 429 "Too Many Requests" errors.

I though of an approache to solve this, have a look:
1.  **Introduce delays:** Modify the test to add a small delay between checking each link to avoid overwhelming GitHub's rate limiter.

btw, could you please check the github actions rate limit? maybe there is some problem there.

smoelius:
@nowak-token I agree with your assessment. But rather than introduce delays, I would rather just skip CHANGELOG.md.

Would you _like to_ submit a PR with that fix? I realize you didn't sign up for this, so I could just as easily submit one myself.

nowak-token:
@smoelius sure, I'll be opening another pr to implement those changes, no worries, I've taken a liking to work on this.

nowak-token:
Ahh the ci passed finally. @smoelius do you think this cab be merged now?

examples/general/abs_home_path/src/lib.rs
```diff
@@ -81,13 +73,8 @@ impl<'tcx> LateLintPass<'tcx> for AbsHomePath {
...
             return;
         }
-
-        if cx
-            .tcx
-            .hir()
-            .parent_iter(expr.hir_id)
-            .any(|(_id, node)| matches!(node, Node::Item(item) if self.is_test_item(item)))
-        {
+        // Skip expressions inside test functions
```
smoelius, line 76:
```suggestion

        // Skip expressions inside test functions
```
Nit (spacing).

examples/general/abs_home_path/src/lib.rs
```diff
@@ -178,52 +124,94 @@ fn ui() {
...
+// Combined test for context allowances (build script and test context)
 #[test]
-fn build_script_allowance() {
+fn context_allowance() {
     use dylint_internal::CommandExt;
     use std::{
         io::{Write, stderr},
         path::Path,
         process::Command,
+        process::Output,
```
smoelius, line 135:
```suggestion
        process::{Command, Output},
```

examples/general/abs_home_path/src/lib.rs
```diff
@@ -178,52 +124,94 @@ fn ui() {
...
     };
 
-    // Skip test if repository is not stored in the user's home directory
+    struct TestCase {
+        manifest_path: &'static str,
+        context_name: &'static str,
```
smoelius, line 140:
```suggestion
        context_name: &'static str,
        manifest_path: &'static str,
```
Let's please put `context_name` first since it's used first in the code below.

examples/general/abs_home_path/src/lib.rs
```diff
@@ -178,52 +124,94 @@ fn ui() {
...
+                let stderr = String::from_utf8_lossy(&output.stderr);
+                // We expect warnings for non-test functions but not for test functions
+                assert!(
+                    stderr.contains("this path might not exist in production"),
+                    "Expected the abs_home_path lint to trigger for non-test functions in test context: {stderr}"
+                );
+                // Verify the test detected non-test functions correctly
+                assert!(
+                    stderr.contains("non_test_function_with_home_path")
+                        || stderr.contains("src/main.rs:9:13"),
```
smoelius, line 173:
```suggestion
                    stderr.contains("src/main.rs:9:13"),
```

examples/general/abs_home_path/src/lib.rs
```diff
@@ -178,52 +124,94 @@ fn ui() {
...
+                    "Expected warning in non-test function in test context: {stderr}"
+                );
+                // Verify that test functions weren't warned about
+                assert!(
+                    !stderr.contains("test_function_with_home_path")
+                        && !stderr.contains("src/main.rs:15:13"),
+                    "The abs_home_path lint should not trigger in test functions: {stderr}"
+                );
```
smoelius, line 181: This feels fragile to me. If the test name changes, or lines are added to or removed from the file, the check will succeed.

I think I would prefer a check something like this:
```rs
stderr.lines().all(|line| !line.contains("src/main.rs") || line.contains("src/main.rs:9:13"))
```
In other words, the only line that is flagged is the one you already checked for.

examples/general/abs_home_path/src/lib.rs
```diff
@@ -178,52 +124,94 @@ fn ui() {
...
-        output.status.success(),
-        "Cargo dylint check failed: {stderr}"
-    );
-    assert!(
-        !stderr.contains("this path might not exist in production"),
-        "The abs_home_path lint was incorrectly triggered in a build script: {stderr}"
-    );
+    for case in test_cases {
+        println!("Testing {} allowance...", case.context_name);
+        let output = Command::new(&cargo_dylint)
```
smoelius, line 203:
```suggestion

        let output = Command::new(&cargo_dylint)
```
Nit (spacing again).

examples/general/abs_home_path/src/lib.rs
```diff
@@ -178,52 +124,94 @@ fn ui() {
...
         process::Command,
+        process::Output,
     };
 
-    // Skip test if repository is not stored in the user's home directory
+    struct TestCase {
+        manifest_path: &'static str,
+        context_name: &'static str,
+        assert_fn: fn(&Output),
+    }
```
smoelius, line 142: I think `TestCase` needs an additional field to determine whether the following should apply:
https://github.com/trailofbits/dylint/blob/ae845edfa22b582bf5d568e622db3607080d706e/examples/general/abs_home_path/src/lib.rs#L208
Note that the original `build_script_allowance` test applied this.

smoelius (commented):
@nowak-token I have a few more changes to request. I think we're almost there though. Thanks very much for continuing to work on this.

nowak-token:
@smoelius Added your suggestions, also made a slight change on how `deny warning` is implemented 
Let me know if this is alright or if I should just make it like the previous one.
https://github.com/trailofbits/dylint/blob/ae845edfa22b582bf5d568e622db3607080d706e/examples/general/abs_home_path/src/lib.rs#L208

smoelius (commented):
What did you mean when you wrote this?
> also made a slight change on how deny warning is implemented

What you have is about what I expected.

Please let me know if you disagree with anything I wrote below, or if you think it doesn't make sense.

nowak-token:
@smoelius Thanks for the suggestions. I'll be traveling this week, hopefully I'll be back by Sunday. Is it alright if I hold off on this till I get back?

smoelius:
> @smoelius Thanks for the suggestions. I'll be traveling this week, hopefully I'll be back by Sunday. Is it alright if I hold off on this till I get back?

Absolutely. I really appreciate you working on this. 🙏

nowak-token:
> It appears the test is hitting GitHub's rate limits when checking all the links (especially commit links) in `CHANGELOG.md`, resulting in numerous HTTP 429 "Too Many Requests" errors.
> 
> I though of an approache to solve this, have a look:
> 
> 1. **Introduce delays:** Modify the test to add a small delay between checking each link to avoid overwhelming GitHub's rate limiter.
> 
> btw, could you please check the github actions rate limit? maybe there is some problem there.

@smoelius it seems the markdown_link_check test is failing again with a 429 error, but this time the error seems to be with, 

>  24 links checked.
  [✖] https://github.com/trailofbits/dylint/tree/master/examples/general/non_local_effect_before_error_return/src/lib.rs → Status: 429
  
 any idea as to why this is happening?

smoelius:
> any idea as to why this is happening?

No, and I'm seeing the same error on master, so it's not your PR. I'll look at this more closely later on today.

nowak-token:
@smoelius I fixed the issue locally by modifying the test to accept 429 (Too Many Requests) status codes in the `markdown_link_check` function in [ci.rs](https://github.com/trailofbits/dylint/blob/master/cargo-dylint/tests/ci.rs#L589). Though I wouldn't call this a fix.
```rs
                .all(|line| {
                    line.is_empty()
                        || line.ends_with(" → Status: 500")
                        || line.ends_with(" → Status: 429")
                }), 
```
 However, I found a better solution:

1. Adding GitHub token support to your link checker configuration:
   - Create a GitHub token with read-only permissions
   - Update your `markdown_link_check.json` config to include:
     ```json
     "httpHeaders": [
       {
         "urls": ["https://github.com/", "https://api.github.com/"],
         "headers": {
           "Authorization": "token ${GITHUB_TOKEN}"
         }
       }
     ]
     ```

2. Set the token as an environment variable in CI:
   - Add `GITHUB_TOKEN` as a repository secret
   - Pass it to your test environment

This will increase your rate limits from 60 to 5000 requests per hour, preventing the 429 errors completely rather than just ignoring them. Found this while going through [GitHub Docs here](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#:~:text=The%20primary%20rate%20limit,5%2C000%20requests%20per%20hour.).

smoelius:
> However, I found a better solution:
> 
> 1. Adding GitHub token support to your link checker configuration: ...

Hmm... A part of me says this is the "right" solution.

But that would only fix the problem in CI. Users running the tests locally would also need a GitHub token.

I guess another option could be to skip the test when not in CI.

Please let me think on this for one day. If you have additional thoughts, you are welcome to share.

smoelius:
OK. I think I'm going to go with your idea, but with something like this:
```rs
if var("GITHUB_TOKEN").is_ok() {
    #[allow(clippy::explicit_write)]
    writeln!(
        stderr(),
        "Skipping `markdown_link_check` test as `GITHUB_TOKEN` is not set"
    )
    .unwrap();
    return;
}
```
That way, it will run CI, and a user can run it locally so long as they set the `GITHUB_TOKEN` environment variable.

Do you want to submit a PR so that you can get credit for this? Or are you happy to let me do it?

nowak-token:
@smoelius Thank you for your suggestion - I like your approach of skipping the test when `GITHUB_TOKEN` isn't set. This is a good compromise that will work for both CI and local testing environments. I'd be happy to implement this solution in a PR. I'll work on it right away, following your guidance with the conditional skip based on the environment variable. This seems more elegant than just accepting 429 errors.
I appreciate you giving me the opportunity to contribute to this fix. I'll submit the changes soon.

smoelius:
@nowak-token Sounds good. Thank you. 🙏

nowak-token:
@smoelius I've already made the changes in #1622. Please go over it when you're free.

smoelius:
@nowak-token Could you rebase onto master? I think the `markdown-link-check` test should pass now.

smoelius:
Thanks, @nowak-token. Sorry #1622 ha become such a bear.

PR #1606, opened 2025-04-28
Fix: rust-toolchain-message

suratkhan:


suratkhan:
@smoelius here is the PR for the rust-toolchain error message fix

smoelius:
Thanks, @suratkhan!

PR #1612, opened 2025-04-29
Skip CHANGELOG.md in markdown link check to avoid GitHub rate limits

nowak-token:
Removed the Changelog link test from the markdown_link_check ci test as mentioned in #1604 
I tested it locally and the test passed locally


nowak-token:
@smoelius check if this can be merged now

smoelius:
Thanks, @nowak-token. Exactly the change I would have made. :)

PR #1637, opened 2025-05-14
Replace unnecessary use of DocumentMut in examples/src/lib.rs

### LINKED ISSUE #1635: Eliminate use of `toml_edit::DocumentMut` where possible
There are several uses of `toml_edit::DocumentMut` that appear unnecessary. Example: https://github.com/trailofbits/dylint/blob/a331d93b275100d5162fa40e0da3a5ac47743127/examples/src/lib.rs#L106-L113
In its place, a `toml::Table` could probably be used.

The task is to perform such a replacement wherever possible.

I'm pretty sure there are a few places where it is not possible, though.

finnfisher-de:
## Description

This PR replaces the use of `toml_edit::DocumentMut` with `toml::Table` in the `examples_do_not_require_rust_src` function in `examples/src/lib.rs`. This change is appropriate since the function only reads values from the TOML file and doesn't need to modify it.

## Changes

- Updated imports to use `toml::Table` directly instead of just `toml`
- Replaced `DocumentMut` with `Table` in the function
- Updated the parsing method to use `Table` directly

## Testing

Verified that ci tests continue to pass with this change.

Fixes #1635


finnfisher-de:
@smoelius please review.

examples/src/lib.rs
```diff
@@ -123,18 +123,13 @@ mod test {
...
-                .and_then(|table| table.get("components"))
-                .and_then(Item::as_array)
-                .unwrap();
-            let components = array
-                .iter()
-                .map(Value::as_str)
-                .collect::<Option<Vec<_>>>()
+                .and_then(|value| value.as_table())
+                .and_then(|toolchain| toolchain.get("components"))
+                .and_then(|value| value.as_array())
```
smoelius, line 131:
```suggestion
                .and_then(|value| value.as_array()).unwrap();
```

examples/src/lib.rs
```diff
@@ -123,18 +123,13 @@ mod test {
...
+                .map(|array| array.iter().filter_map(|v| v.as_str()).collect::<Vec<_>>())
```
smoelius, line 132:
```suggestion
            let components = array
                .iter()
                .map(Value::as_str)
                .collect::<Option<Vec<_>>>()
                .unwrap();
```
I think I extracted the components separately on purpose. The reason is: if `as_str` returns `None` for any value in the array, then the overall result will be `None` and the `unwrap` will fail (indicating that there is a problem).

Note that the `filter_map` approach would silently ignore `None`s returned `as_str`.

I think I prefer the former, loud and in-your-face approach.

smoelius (commented):
Thanks, @finnfisher-de. I have just a few comments below. After you've had a chance to review them, I think this can be merged.

I think there are a few more uses of `DocumentMut` that can be changed to `toml::Table`. Are you interested in tackling those as well?

Here are the ones I noticed:
- https://github.com/trailofbits/dylint/blob/664b774895cf005ceae4b86d3c6b027d01b61b32/dylint-link/src/main.rs#L48
- https://github.com/trailofbits/dylint/blob/664b774895cf005ceae4b86d3c6b027d01b61b32/internal/src/packaging.rs#L158

internal/src/clippy_utils.rs
```diff
@@ -4,6 +4,7 @@ use std::{
     fs::{read_to_string, write},
     path::Path,
 };
+use toml;
```
smoelius, line 7:
```suggestion
```
Hmm. Here, too, a lint should have fired.

I hope you don't mind. I'm going to spend a day or two looking into this.

smoelius: This lint specifically should have fired: https://rust-lang.github.io/rust-clippy/master/index.html#/single_component_path_imports

smoelius:
@finnfisher-de Is this good to go in your opinion?

finnfisher-de:
@smoelius LGTM 🚀

smoelius:
@finnfisher-de Could you please squash your commits down to one?

smoelius:
Thanks, @finnfisher-de!

internal/src/clippy_utils.rs
smoelius: Leaving a note here for myself.

The reason for `single_component_path_imports` not firing seems related to `clippy_utils` being a public module: https://github.com/trailofbits/dylint/blob/6db435a98a7045fe64ce44a4c7901f56016c8c81/internal/src/lib.rs#L7-L8

If you remove the `pub`, re-add the `use toml;`, and run the following commands, a warning is generated:
```sh
cd internal
cargo clippy --all-features
```

PR #1671, opened 2025-06-07
Sanitize environment in `package_with_root`

smoelius:
A problem arises when the linted project uses edition 2021 but the lints uses edition 2024.

If the environment is not sanitized, the edition 2021 tools are propagated to the lints via environment variables, and the lints fail to build. What's more, the errors are suppressed, and the generic "No libraries were found." is printed.

~This change needs a test.~


smoelius:
Thanks to @fcasal for providing the repo that demonstrated this bug.

PR #1802, opened 2025-10-30
Update `README.md` to include examples of supported git revision fields

brannondorsey:
This updates the README to document that `workspace.metadata.dylint.libraries` can include git repositories pinned to specific tags, branches, and revisions.

This is important, especially when lints are run via CI, to prevent changes in this repo from breaking automated builds.

PR #1803, opened 2025-10-30
Update `README.md` to suggest installing binary crates with `--locked`

brannondorsey:
Update install instructions to suggest installing `cargo-dylint` and `dylint-link` with the their published lockfiles.

The only reason I can think of why you wouldn't want to do this would be if lockfiles aren't published or kept up to date with these packages. But I can see that `Cargo.lock` is in fact checked published with both of these binary crates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 Run Rust lints from dynamic libraries (EuroRust 2024 [slides] and [video])
 
 ```sh
-cargo install cargo-dylint dylint-link
+cargo install cargo-dylint dylint-link --locked
 ```
 
 Dylint is a Rust linting tool, similar to Clippy. But whereas Clippy runs a predetermined, static set of lints, Dylint runs lints from user-specified, dynamic libraries. Thus, Dylint allows developers to maintain their own personal lint collections.
@@ -33,7 +33,7 @@ The next two steps install Dylint and run all of this repository's [general-purp
 1. Install `cargo-dylint` and `dylint-link`:
 
    ```sh
-   cargo install cargo-dylint dylint-link
+   cargo install cargo-dylint dylint-link --locked
    ```
 
 2. Run `cargo-dylint`:


smoelius:
> Update install instructions to suggest installing `cargo-dylint` and `dylint-link` with the their published lockfiles.
> 
> The only reason I can think of why you wouldn't want to do this would be if lockfiles aren't published or kept up to date with these packages. But I can see that `Cargo.lock` is in fact checked published with both of these binary crates.

I appreciate the PR, @brannondorsey. But I feel like this is an issue on which there is not consensus. For example, one could argue that by not using `--locked`, one gets bugfixes that did not exist when the package was published. Do you not agree?

brannondorsey:
> > Update install instructions to suggest installing `cargo-dylint` and `dylint-link` with the their published lockfiles.
> > The only reason I can think of why you wouldn't want to do this would be if lockfiles aren't published or kept up to date with these packages. But I can see that `Cargo.lock` is in fact checked published with both of these binary crates.
> 
> I appreciate the PR, @brannondorsey. But I feel like this is an issue on which there is not consensus. For example, one could argue that by not using `--locked`, one gets bugfixes that did not exist when the package was published. Do you not agree?

@smoelius This is a fair comment. Here's my perspective on it:

For getting started instructions, you want to provide an experience that is as friction-less as possible. The best way to do that, IMO, is to suggest installing using the same exact dependencies that the chart maintainers last published with (because their release tests presumably tested against that exact same set of dependencies before release, and the tests passed).

Were this package to go unmaintained for a period of time, the chances that an install with `--locked` works remain fixed, while the chances that an install _without_ it starts breaking, increases.

On a number of occasions, I've seen third party library crates publish breaking changes in non-semver breaking version bumps (because they didn't have good test/usage coverage). Including `--locked` prevents this behavior from impacting your downstream users.

Another thing I like about throwing a `--locked` in there is that the user can consciously choose to omit it, rather than forgetting that feature exists to begin with.

smoelius:
Your points are all valid. But I worry that adding `--locked` would make it look like Dylint was taking a stand on this issue, and I would rather it not be perceived that way.

Also, Dylint is actually tested quite frequently. First, there are the PRs to this repo. But also several other repos install `cargo-dylint` without `--locked`. So if a problem were to arise, I think it would be uncovered relatively quickly.

So I am going to close this. But please don't interpret that as my disinterest, because I am genuinely, truly appreciative of the PR! 🙏

And we can continue to discuss if you like.

brannondorsey:
@smoelius no problem. :smile_cat: 

Let me know if there is anything that still needs to be done to get https://github.com/trailofbits/dylint/pull/1802 merged.

smoelius:
> @smoelius no problem. 😸
> 
> Let me know if there is anything that still needs to be done to get #1802 merged.

All good. Thanks, @brannondorsey.

PR #1847, opened 2025-11-25
Fix auto correct non determinism

### LINKED ISSUE #1669: `cargo dylint upgrade --auto-correct` is non-deterministic
Example: commits from four PRs to upgrade the lints to nightly-2025-05-14:
- https://github.com/trailofbits/dylint/pull/1646/commits/4182b143d2dc865d6e8d615024d264c8d0fa1fb3?file-filters%5B%5D=.rs&show-viewed-files=true
- https://github.com/trailofbits/dylint/pull/1656/commits/e769aa92951e0d335cb243685191762549b3ae9f?file-filters%5B%5D=.rs&show-viewed-files=true
- https://github.com/trailofbits/dylint/pull/1660/commits/319f33fa3a3f8476b1f8b6d94974659b120737b1?file-filters%5B%5D=.rs&show-viewed-files=true
- https://github.com/trailofbits/dylint/pull/1668/commits/146663c299ad153a8f3b0e9763b96a2007e42168?file-filters%5B%5D=.rs&show-viewed-files=true

james-miller-93:
Fix non-deterministic behavior in auto-correct by using BTreeMap

Replaces HashMap<Rewrite, Oid> with BTreeMap<Rewrite, Oid> throughout the auto-correct module to ensure deterministic iteration order.

The root cause of issue #1669  was that HashMap iteration order is non-deterministic in Rust (uses a randomized hasher). When multiple rewrites produced the same replacement text with the same score but different Git commit OIDs, the iteration order determined which OID got selected, leading to inconsistent results across runs.

Changes:
  - Add Ord and PartialOrd trait derivations to Rewrite struct
  - Replace HashMap with BTreeMap in collect_rewrites() return type
  - Replace HashMap with BTreeMap in applicable_rewrites() parameter
  - Update imports to use BTreeMap instead of HashMap

This ensures that cargo dylint upgrade --auto-correct produces consistent, reproducible results on every run.

Fixes #1669

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/dylint/src/package_options/auto_correct/mod.rs
+++ b/dylint/src/package_options/auto_correct/mod.rs
@@ -6,7 +6,7 @@ use anyhow::{Context, Result};
 use dylint_internal::{clippy_utils::clippy_repository, git2::Oid};
 use rewriter::{LineColumn, Rewriter, Span, interface::Span as _};
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::BTreeMap,
     env::current_dir,
     fs::{read_to_string, write},
     ops::Range,
@@ -180,7 +180,7 @@ pub fn auto_correct_revertible(
 }
 
 fn applicable_rewrites<'rewrite>(
-    rewrites: &'rewrite HashMap<Rewrite, Oid>,
+    rewrites: &'rewrite BTreeMap<Rewrite, Oid>,
     highlight: &Highlight,
 ) -> Result<ReplacementSourceMap<'rewrite>> {
     let mut replacement_source_map = ReplacementSourceMap::new();
--- a/dylint/src/package_options/auto_correct/rewrite/mod.rs
+++ b/dylint/src/package_options/auto_correct/rewrite/mod.rs
@@ -5,14 +5,14 @@ use dylint_internal::{
     env,
     git2::{Commit, Diff, DiffHunk, Oid, Patch, Repository},
 };
-use std::{cmp::min, collections::HashMap, ops::Range, time::Instant};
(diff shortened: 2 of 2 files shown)


smoelius:
> The root cause of issue #1669 was that HashMap iteration order is non-deterministic in Rust (uses a randomized hasher).

Thanks very much for working on this, @james-miller-93. I think Claude got this one wrong, though. I think the root cause is that the Rust compiler uses multiple threads, and that when one fails, the compiler waits for the other threads to finish. So the more threads one compiles with, the more rewrites one is likely to get out.

What makes this more complicated is that Cargo (the build tool) not rustc (the compiler) applies the fixes. To the best of my knowledge, there is no way to control the number of threads that Cargo uses. So, I don't see an easy way to fix this.

I have been working on a [PR](https://github.com/trailofbits/dylint/pull/1786) to expand the number of auto-correct-related tests. The expanded set of tests tries to verify that the test output [is a subset](https://github.com/trailofbits/dylint/pull/1786/files#diff-d6172a2027b70158a5804ab6123f7289469726f5be466d7f3e75cf41ebe0abd3R231) of what is expected. Note that this doesn't actually fix anything. It just provides a tiny bit more assurance that things are working correctly.

I apologize that the PR I linked has no description. I'll take that as a note to better document my PRs.

Please let me know if anything I've written doesn't sound like it makes sense.

smoelius:
@james-miller-93 I merged https://github.com/trailofbits/dylint/pull/1786, which I linked above. Thanks again for working on this. 🙏

PR #1877, opened 2025-12-13
Update upgrade_examples.yml to not use secret token

smoelius:


smoelius:
GitHub documentation suggests it may be impossible to trigger subsequent workflows with the default GitHub token.

See #1878, for example, which was generated from this workflow. It did not trigger the CI checks.

Closing.

PR #1888, opened 2025-12-22
Add `match_hir` library and `disallowed_pattern` lint

smoelius:


smoelius:
Could this be made into a proc macro?

The benefit would be that if I got tired of maintaining this library, one could just `cargo expand` the code and use that, i.e., the code that used this library would not be indelibly lost.

PR #1972, opened 2026-05-14
[ci] Build Linux release binaries

### LINKED ISSUE #1971: Support `cargo-binstall`
## Overview

Currently, no binaries are published along with releases, forcing CI pipelines to manually build and cache the `cargo-dylint` and `dylint-link` binaries. This prevents the use of [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall/) / [`taiki-e/install-action`](https://github.com/taiki-e/install-action). It would be great if we could add a workflow that built binaries for `x86_64-unknown-linux-gnu` and `aarch64-unknown-linux-gnu` + published them along with the release when a tag is pushed.

Would be willing to open a PR if the maintainers are open to it.

clabby:
## Overview

Adds a step to the release workflow to build, sign, and publish the `cargo-dylint` and `dylint-link` binaries.

### Signing

The signing step uses the [`JIT Signing strategy`](https://github.com/cargo-bins/cargo-binstall/blob/bbd0a7de0e936c8c626fa0e93584169d58ff1307/SIGNING.md#just-in-time-signing) described in the `cargo-binstall` docs, creating an ephemeral key and signing each binary before uploading.

`cargo-binstall` will verify the `*.sig` files against the public key included in the crate's manifest for the given version, which is injected prior to publishing the crates.

closes #1971

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,9 +6,63 @@ on:
       - "v*"
 
 jobs:
+  build-binaries:
+    name: Build binaries (${{ matrix.target }})
+    runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+          - os: ubuntu-24.04-arm
+            target: aarch64-unknown-linux-gnu
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ matrix.target }}
+
+      - name: Build binaries
+        run: cargo build --locked --release --target ${{ matrix.target }} -p cargo-dylint -p dylint-link
+
+      - name: Package binaries
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          VERSION="${GITHUB_REF_NAME#v}"
+          BINARIES="cargo-dylint dylint-link"
+
+          for BINARY in $BINARIES; do
+            PACKAGE_NAME="${BINARY}-${{ matrix.target }}-v${VERSION}"
+            mkdir -p "${PACKAGE_NAME}"
+            cp "target/${{ matrix.target }}/release/${BINARY}" "${PACKAGE_NAME}/"
+
+            ARCHIVE="${PACKAGE_NAME}.tar.gz"
+            tar -czf "${ARCHIVE}" "${PACKAGE_NAME}"
+            sha256sum "${ARCHIVE}" > "${ARCHIVE}.sha256"
+          done
+
+      - name: Upload binary artifacts
+        uses: actions/upload-artifact@v7
+        with:
+          name: binaries-${{ matrix.target }}
+          path: |
+            cargo-dylint-${{ matrix.target }}-v*.tar.gz
+            cargo-dylint-${{ matrix.target }}-v*.tar.gz.sha256
+            dylint-link-${{ matrix.target }}-v*.tar.gz
+            dylint-link-${{ matrix.target }}-v*.tar.gz.sha256
+          if-no-files-found: error
+
   publish:
+    needs: build-binaries
     runs-on: ubuntu-latest
     permissions:
+      actions: read
+      contents: read
       id-token: write
     steps:
       - uses: actions/checkout@v6
@@ -21,12 +75,48 @@ jobs:
       - uses: rust-lang/crates-io-auth-action@v1
         id: auth
 
+      - name: Download binary artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: dist
+          merge-multiple: true
+
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: rsign2
+          fallback: cargo-install
+
+      - name: Sign binary artifacts
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          MANIFESTS="cargo-dylint/Cargo.toml dylint-link/Cargo.toml"
+          trap 'rm -f minisign.key minisign.pub' EXIT
+
+          rsign generate -f -W -p minisign.pub -s minisign.key
+          echo "::add-mask::$(tail -n 1 minisign.key)"
+
+          PUBLIC_KEY="$(tail -n 1 minisign.pub)"
+
+          for MANIFEST in $MANIFESTS; do
+            printf '\n[package.metadata.binstall.signing]\n' >> "${MANIFEST}"
+            printf 'algorithm = "minisign"\n' >> "${MANIFEST}"
+            printf 'pubkey = "%s"\n' "${PUBLIC_KEY}" >> "${MANIFEST}"
+          done
+
+          for ARCHIVE in dist/*.tar.gz; do
+            rsign sign -W -s minisign.key -x "${ARCHIVE}.sig" "${ARCHIVE}"
+          done
+
       - name: Publish
         run: |
+          set -euo pipefail
+
           # smoelius: Publishing in this order ensures that all dependencies are met.
           DIRS="internal driver dylint-link dylint cargo-dylint utils/linting utils/testing"
           for DIR in $DIRS; do
-              pushd "$DIR" && (cargo publish || true) && popd
+              cargo publish --allow-dirty --manifest-path "$DIR/Cargo.toml"
           done
         env:
           CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}
@@ -44,6 +134,10 @@ jobs:
           tag_name: ${{ github.ref }}
           name: Release ${{ steps.get-version.outputs.version }}
           body_path: body.md
+          files: |
+            dist/*.tar.gz
+            dist/*.sha256
+            dist/*.sig
           draft: false
           prerelease: ${{ contains(github.ref, 'pre') || contains(github.ref, 'rc') }}
           token: ${{ secrets.REPO_TOKEN }}


smoelius:
Thanks, @clabby. I will try to look at this this weekend.

.github/workflows/release.yml
```diff
@@ -6,9 +6,84 @@ on:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+            use_cross: false
+          - os: ubuntu-latest
+            target: aarch64-unknown-linux-gnu
+            use_cross: true
```
smoelius: My colleague points out that GitHub provides [Ubuntu ARM64 runners](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md) now. Could you use that instead of `cross`? Or is there a reason to prefer `cross`?

.github/workflows/release.yml
```diff
@@ -6,9 +6,84 @@ on:
...
+          VERSION="${GITHUB_REF_NAME#v}"
+          BINARIES="cargo-dylint dylint-link"
+
+          checksum() {
+            if command -v shasum >/dev/null 2>&1; then
+              shasum -a 256 "$1" > "$1.sha256"
+            else
+              sha256sum "$1" > "$1.sha256"
+            fi
+          }
```
smoelius: Is there a reason this can't be just `sha256sum`?

.github/workflows/release.yml
```diff
@@ -21,12 +96,49 @@ jobs:
...
 
+      - name: Download binary artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: dist
+          merge-multiple: true
+
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: rsign2
```
smoelius, line 86: You are not running `minisign` at all, correct?

Because I am not quite following the flow here.

What does the following line do in the Cargo.toml?
```toml
algorithm = "minisign"
```

.github/workflows/release.yml
```diff
@@ -6,9 +6,84 @@ on:
...
+            mkdir -p "${PACKAGE_NAME}"
+            cp "target/${{ matrix.target }}/release/${BINARY}" "${PACKAGE_NAME}/"
+
+            ARCHIVE="${PACKAGE_NAME}.tar.gz"
+            tar -czf "${ARCHIVE}" "${PACKAGE_NAME}"
+            checksum "${ARCHIVE}"
+          done
+
+      - name: Upload binary artifacts
+        uses: actions/upload-artifact@v4
```
smoelius:
```suggestion
        uses: actions/upload-artifact@v7
```

.github/workflows/release.yml
```diff
@@ -21,12 +96,49 @@ jobs:
...
+        with:
+          tool: rsign2
+          fallback: cargo-install
+
+      - name: Sign binary artifacts
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          MINISIGN_KEY=minisign.key
```
smoelius: What is the advantage to defining `MINISIGN_KEY`?

smoelius (commented):
Overall, I think this looks really good!

I asked my colleague for feedback, and he provided some that I included below.

I presume you haven't tested this, correct?

.github/workflows/release.yml
clabby: We can - either works! I'll update to use the arm runner.

.github/workflows/release.yml
clabby: In the past, GitHub has broken their images (i.e., not installing `coreutils` from time to time.)

Right now it looks like the `ubuntu-latest` image has `sha256sum`, though this is mainly a fallback to keep things working smoothly.

.github/workflows/release.yml
clabby: We are running minisign - rsign2 is an implementation of minisign used by `cargo-binstall`: https://github.com/cargo-bins/cargo-binstall/blob/7c6d6446dab5843b9ab70840f15fbad89e729fbd/.github/workflows/release-cli.yml#L41-L43

> What does the following line do in the Cargo.toml?

This is just a required field for forward compatibility with new signing schemes, it seems: https://github.com/cargo-bins/cargo-binstall/blob/7c6d6446dab5843b9ab70840f15fbad89e729fbd/crates/binstalk-types/src/cargo_toml_binstall.rs#L225-L226. `cargo-binstall` currently only supports minisign: https://github.com/cargo-bins/cargo-binstall/blob/7c6d6446dab5843b9ab70840f15fbad89e729fbd/crates/binstalk-types/src/cargo_toml_binstall.rs#L240-L241

.github/workflows/release.yml
clabby: Just to avoid repetition (though on second read, might also be nice to add a variable for the public key file.)

.github/workflows/release.yml
clabby: https://github.com/trailofbits/dylint/pull/1972/commits/449d9ea75999e862c6c7e3a9d011880fdf39a3f7

.github/workflows/release.yml
clabby: https://github.com/trailofbits/dylint/pull/1972/commits/56ab746cc0dc6daad780684c0db5aba808bb3471

.github/workflows/release.yml
clabby: https://github.com/trailofbits/dylint/pull/1972/commits/cc290bb404f624276f5bd624fd7ccb74cf8b2e40

clabby:
Tested the release workflow (sans `crates.io` publishing) @ https://github.com/clabby/dylint/actions/runs/26045748976 - release artifact: https://github.com/clabby/dylint/releases/tag/v6.1.3

.github/workflows/release.yml
smoelius: I understand your reasoning, but let's please just use `sha256sum`.

.github/workflows/release.yml
smoelius: So if `rsign2` was not installed, would this break?

.github/workflows/release.yml
smoelius: Actually, could you please remove `MINISIGN_PRIV_KEY` and `MINISIGN_PUB_KEY`?

I think I would prefer to not have them.

smoelius:
> Tested the release workflow (sans `crates.io` publishing) @ https://github.com/clabby/dylint/actions/runs/26045748976 - release artifact: https://github.com/clabby/dylint/releases/tag/v6.1.3

Nice!

Should I be able to `cargo-binstall` those, though?

I did the below experiment and it fell back to `cargo-quickinstall`. Is that what you would expect?

```sh
$ cargo binstall --only-signed cargo-dylint --git https://github.com/clabby/dylint --version 6.1.3
 INFO resolve: Resolving package: 'cargo-dylint@=6.1.3'
 INFO Cloning with worktree::receiving pack: Enumerating objects → 664.0 objects
 INFO ℹCloning with worktree::receiving pack::create index file::indexing → done 664.0 objects in 0.14s (4.6k objects/s)
 INFO ℹCloning with worktree::receiving pack::create index file::decompressing → done 4.0MB in 0.14s (27.5MB/s)
 INFO ℹCloning with worktree::receiving pack::create index file::Resolving → done 664.0 objects in 0.06s (12.0k objects/s)
 INFO ℹCloning with worktree::receiving pack::create index file::Decoding → done 4.8MB in 0.06s (86.4MB/s)
 INFO ✓Cloning with worktree::receiving pack::create index file::writing index file → done 19.7kB in 0.00s (115.0MB/s)
 INFO ✓Cloning with worktree::receiving pack::create index file → done 664.0 objects in 0.20s (3.3k objects/s)
 INFO ℹCloning with worktree::receiving pack::read pack → done 2.4MB in 0.27s (9.0MB/s)
 INFO ℹCloning with worktree::receiving pack::checkout → done 450.0 files in 0.07s (6.4k files/s)
 INFO ℹCloning with worktree::receiving pack::writing → done 4.4MB in 0.07s (63.5MB/s)
 INFO resolve: Verified signature for package 'cargo-dylint-6.0.0-aarch64-apple-darwin': timestamp:1778689584	file:cargo-dylint-6.0.0-aarch64-apple-darwin.tar.gz	hashed
 WARN The package cargo-dylint v6.0.0 (aarch64-apple-darwin) has been downloaded from third-party source QuickInstall
 INFO This will install the following binaries:
 INFO   - cargo-dylint => /Users/samuel.moelius/.cargo/bin/cargo-dylint
Do you wish to continue? [yes]/no yes
 INFO Installing binaries...
 INFO Done in 9.319703625s
```

.github/workflows/release.yml
clabby: https://github.com/trailofbits/dylint/pull/1972/changes/10716bfdc9442e07d0e2a9f139e5682c2fe24efd

.github/workflows/release.yml
clabby: https://github.com/trailofbits/dylint/pull/1972/changes/85dcfce0e4e12f2b9f329f89490793e7112460fa

.github/workflows/release.yml
clabby: `rsign2` is only required for signing within the release workflow, though `cargo-binstall` embeds a signature verifier into its binary using the [`minisign_verify`](https://docs.rs/minisign-verify/latest/minisign_verify/) crate: https://github.com/cargo-bins/cargo-binstall/blob/7c6d6446dab5843b9ab70840f15fbad89e729fbd/crates/binstalk-fetchers/src/signing.rs#L39

clabby:
> Should I be able to `cargo-binstall` those, though?
> 
> I did the below experiment and it fell back to `cargo-quickinstall`. Is that what you would expect?

This is what I'd expect from a quick read - it does appear to fall back per the docs. The `Cargo.toml` in my fork still points to `trailofbits/dylint` in its crate metadata, so it falls back to `cargo-quickinstall` as a secondary strategy to deal with the third-party artifact host.

[From `cargo-binstall`'s documentation](https://crates.io/crates/cargo-binstall):

> Binstall works by fetching the crate information from crates.io and searching the linked repository for matching releases and artifacts, falling back to the [quickinstall](https://github.com/alsuren/cargo-quickinstall) third-party artifact host

`cargo-binstall` seems to try the strategies listed [here](https://github.com/cargo-bins/cargo-binstall/blob/7c6d6446dab5843b9ab70840f15fbad89e729fbd/crates/binstalk-types/src/cargo_toml_binstall.rs#L44-L52) in-order, assuming none have been [manually disabled](https://github.com/cargo-bins/cargo-binstall/blob/7c6d6446dab5843b9ab70840f15fbad89e729fbd/crates/bin/src/args.rs#L207-L221).

.github/workflows/release.yml
```diff
@@ -6,9 +6,64 @@ on:
       - "v*"
 
 jobs:
+  build-binaries:
+    name: Build binaries (${{ matrix.target }})
+    runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
```
smoelius:
```suggestion
```
`fail-fast: false` is not necessary, right?

.github/workflows/release.yml
```diff
@@ -21,12 +76,48 @@ jobs:
...
+
       - name: Publish
         run: |
+          set -euo pipefail
+
           # smoelius: Publishing in this order ensures that all dependencies are met.
           DIRS="internal driver dylint-link dylint cargo-dylint utils/linting utils/testing"
           for DIR in $DIRS; do
-              pushd "$DIR" && (cargo publish || true) && popd
+              (cd "$DIR" && cargo publish --allow-dirty)
```
smoelius:
```suggestion
              cd "$DIR" && cargo publish --allow-dirty
```
Nit.

smoelius (commented):
After making these changes, please squash your commits down to one, and then I will merge.

.github/workflows/release.yml
clabby: These parenthesis are required to run `cd` in a subshell. I instead updated to

```sh
cargo publish --allow-dirty --manifest-path "$DIR/Cargo.toml"
```

smoelius:
Thanks, @clabby!

smoelius:
@clabby Are you waiting for signed binaries to be published?

Issue #1970, opened 2026-05-11
Failures building Dylint driver from crates.io


smoelius:
Recently, I started seeing failures like the one at the end of this issue.

The bug was fixed in commit https://github.com/trailofbits/dylint/commit/b861ff90351aded1a07b1751468d3fe15fc61a90, but it was not published to crates.io.

I should consider installing the driver from GitHub rather than crates.io, so that bugs like this are fixed immediately.

```
     Compiling dylint_driver v5.0.0
  error[E0609]: no field `env_depinfo` on type `&rustc_session::parse::ParseSess`
     --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/dylint_driver-5.0.0/src/lib.rs:251:31
      |
  251 |             sess.parse_sess().env_depinfo.lock().insert((
      |                               ^^^^^^^^^^^ unknown field
      |
      = note: available fields are: `unstable_features`, `config`, `check_config`, `edition`, `raw_identifier_spans` ... and 7 others
  
  error[E0609]: no field `env_depinfo` on type `&rustc_session::parse::ParseSess`
     --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/dylint_driver-5.0.0/src/lib.rs:255:31
      |
  255 |             sess.parse_sess().env_depinfo.lock().insert((
      |                               ^^^^^^^^^^^ unknown field
      |
      = note: available fields are: `unstable_features`, `config`, `check_config`, `edition`, `raw_identifier_spans` ... and 7 others
  
  error[E0609]: no field `env_depinfo` on type `&rustc_session::parse::ParseSess`
     --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/dylint_driver-5.0.0/src/lib.rs:259:31
      |
  259 |             sess.parse_sess().env_depinfo.lock().insert((
      |                               ^^^^^^^^^^^ unknown field
      |
      = note: available fields are: `unstable_features`, `config`, `check_config`, `edition`, `raw_identifier_spans` ... and 7 others
  
  error[E0609]: no field `file_depinfo` on type `&rustc_session::parse::ParseSess`
     --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/dylint_driver-5.0.0/src/lib.rs:277:26
      |
  277 |                         .file_depinfo
      |                          ^^^^^^^^^^^^ unknown field
      |
[...]

PR #1985, opened 2026-05-28
fix(release): avoid embedding `dylint-driver` source paths

clabby:
## Overview

Detail that was overlooked in #1971: Release builds of `cargo-dylint` currently compile from the repository checkout, so `dylint/build.rs` records the sibling driver directory as an absolute path. `cargo-binstall` users then receive a binary that tries to build `dylint_driver` from the release runner's `/home/runner/work/...` checkout, which does not exist outside that job.

Add an explicit `DYLINT_DRIVER_FROM_CRATES_IO` build-time envvar and set it in the release workflow. Packaged binaries now generate driver crates that depend on `dylint_driver` by version from crates.io, while local workspace builds continue using the sibling driver path for development.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,6 +23,8 @@ jobs:
 
       - name: Build binaries
         run: cargo build --locked --release --target ${{ matrix.target }} -p cargo-dylint -p dylint-link
+        env:
+          DYLINT_DRIVER_FROM_CRATES_IO: 1
 
       - name: Package binaries
         shell: bash
--- a/dylint/build.rs
+++ b/dylint/build.rs
@@ -1,12 +1,15 @@
 use dylint_internal::{cargo::cargo_home, env};
 use std::{fs::OpenOptions, io::Write, path::Path};
 
+const DYLINT_DRIVER_FROM_CRATES_IO: &str = "DYLINT_DRIVER_FROM_CRATES_IO";
+
 fn main() {
     write_dylint_driver_manifest_dir();
 
     #[cfg(feature = "__cargo_cli")]
     println!("cargo:rustc-cfg=__library_packages");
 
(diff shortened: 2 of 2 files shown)


smoelius:
@clabby Thanks very much for following up on this.

This is related to #1970, which I opened a couple of weeks ago.

I think I would prefer to not rely on external resources when building a driver. That is, if a driver doesn't currently exist, cargo-dylint could unpack a tar file and build from that. It does something similar for library templates now. I think I could implement that change within the next few days.

Having said that, is the problem you describe a blocker? Are you stuck without this change?

clabby:
@smoelius Thanks for the quick response.

We're not blocked without this, though the downstream workaround feels a bit messy. The issue can be circumvented by placing the `dylint` source at the expected `DYLINT_DRIVER_MANIFEST_DIR`: https://github.com/commonwarexyz/monorepo/pull/3903/changes#r3318545990.

Feel free to take an alternative approach - mainly, what I'd like is for the `cargo-dylint` binary to behave exactly as it does if built from the source published to `crates.io` with `cargo install cargo-dylint` as it does when we use `cargo binstall`. This PR makes a minimal adjustment to do that by setting `DYLINT_DRIVER_MANIFEST_DIR` to `None` when installing from `crates.io` or the published release artifacts.

smoelius:
@clabby I am really sorry for having sat on this for so long.

I tried an approach where the driver source code is embedded in `cargo-dylint` (and I think that should be the long-term solution). But that approach required more design decisions than I anticipated.

So, for now, I adopted an approach very similar to yours, but with a Cargo configuration instead of an environment variable (#2011 ).

I published a [release](https://github.com/trailofbits/dylint/releases/tag/v6.0.2) with the fix, and `cargo binstall cargo-dylint` installs an executable that seems to work for me (e.g., it seems to successfully build a driver).

I'd be happy to hear your feedback (aside: I may be AFK for a bit). But, regardless, I hope the new release meets your needs.

Thanks again for implementing this binary publishing!

PR #2016, opened 2026-07-29
fix: ignore explicitly typed variables in misleading_variable_name

### LINKED ISSUE #1903: `misleading_variable_name` works weirdly
It seems that `misleading_variable_name` starts matching names from the end and only a small part of them. Is this normal behavior? If so, I suggest adding a setting to this lint that allows you to specify which side to start the comparison from.

```rust
let up: UpdateSchemePath = path.into_inner(); // negative
let query: UpdateSchemeQuery = query.into_inner(); 
    ^^^^^
// `dropshot::extractor::query` exports a type `Query`, which is not     
// the type of `query` use a name other than `query`
```

VrtxOmega:
Fixes #1903.

## Summary

- skip `misleading_variable_name` for let bindings with an explicit type annotation
- add a UI regression case proving an explicitly typed binding stays quiet
- document the new fourth condition and regenerate the checked-in example README

## Verification

I locally verified the complete amended diff:

- Before the fix, the explicit `String` binding produced a seventh warning and failed the UI fixture. After the fix, the standalone UI test passes while the six existing warning expectations remain unchanged.
- `PATH="$PWD/target/debug:$PATH" CARGO_TARGET_DIR="$PWD/target" cargo +nightly-2026-05-28 test --manifest-path examples/restriction/misleading_variable_name/Cargo.toml` — 1 UI test passed
- `cargo test -p cargo-dylint --test ci format_example_readmes` — 1 README consistency test passed
- `cargo +nightly-2026-05-28 fmt --all --check`
- `git diff --check`

## Assistance disclosure

OpenAI Codex helped inspect the maintainer feedback and execute the local checks. I reviewed the code change, generated README, diff, and verification output before pushing.

### DIFF
--- a/examples/restriction/misleading_variable_name/README.md
+++ b/examples/restriction/misleading_variable_name/README.md
@@ -2,8 +2,9 @@
 
 ### What it does
 
-Checks for variables satisfying the following three conditions:
+Checks for variables satisfying the following four conditions:
 
+- The variable is declared without an explicit type.
 - The variable is initialized with the result of a function call.
 - The variable's name matches the name of a type defined within the module in which the
   function is defined.
--- a/examples/restriction/misleading_variable_name/src/lib.rs
+++ b/examples/restriction/misleading_variable_name/src/lib.rs
@@ -20,8 +20,9 @@ use std::collections::BTreeMap;
 dylint_linting::declare_late_lint! {
     /// ### What it does
     ///
-    /// Checks for variables satisfying the following three conditions:
+    /// Checks for variables satisfying the following four conditions:
     ///
+    /// - The variable is declared without an explicit type.
     /// - The variable is initialized with the result of a function call.
     /// - The variable's name matches the name of a type defined within the module in which the
     ///   function is defined.
@@ -60,6 +61,7 @@ impl<'tcx> LateLintPass<'tcx> for MisleadingVariableName {
                         kind: PatKind::Binding(_, _, ident, _),
                         ..
                     },
+                ty: None,
                 init: Some(init),
                 ..
             }) = stmt.kind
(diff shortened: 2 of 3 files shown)


smoelius:
Thanks, @VrtxOmega. Please forgive me for pasting the below Claude review, but I think I agree with it. Could you please address it? 

I think the change should be straightforward, so please fold it into the current commit.

Also, please sign the CLA.

## 1. The doc comment and generated README are now inaccurate

[src/lib.rs:24](examples/restriction/misleading_variable_name/src/lib.rs#L24) still reads:

> Checks for variables satisfying the following **three** conditions:

There are now four. The new condition — the variable has no explicit type annotation — is the one users are most likely to need, since it is the documented way to silence a false positive.

[cargo-dylint/tests/ci.rs:343](cargo-dylint/tests/ci.rs#L343) regenerates every example's `README.md` from its lint doc comment and asserts equality, so the doc comment and
[examples/restriction/misleading_variable_name/README.md](examples/restriction/misleading_variable_name/README.md) must be updated together (`BLESS=1 cargo test`). CI passes today only because neither file was touched.

Suggested wording:

```rust
/// Checks for variables satisfying the following four conditions:
///
/// - The variable is declared without an explicit type.
/// - The variable is initialized with the result of a function call.
/// - The variable's name matches the name of a type defined within the module in which the
///   function is defined.
/// - The variable's type is not the matched type.
```

VrtxOmega:
Thanks — addressed in `86edddabe703dc113b10f23d6f24ad5e5cf39769` and folded into the existing commit as requested.

- Updated the rustdoc to describe all four conditions.
- Regenerated the checked-in README through `format_example_readmes`.
- Signed the Trail of Bits CLA; the CLA check is now passing.
- Re-ran the focused UI test, README consistency test, pinned-nightly format check, and `git diff --check`; all pass locally.

The amended CI matrix is running now.

smoelius:
Beautiful. Thank you very much, @VrtxOmega!

PR #2072, opened 2026-09-10
Fix dylint.toml dependency tracking to preserve incremental compilation

### LINKED ISSUE #2015: dylint.toml contents inserted into file_depinfo as a path, breaking incremental compilation
## Summary

`try_init_config_guarded` inserts the **contents** of `dylint.toml` into rustc's `file_depinfo`, which expects a **path**. Every line of the config lands in the crate's `.d` file as a filename that never exists, so Cargo marks every workspace crate stale on every run and incremental compilation is lost.

## Location

[`utils/linting/src/lib.rs:647-651`](https://github.com/trailofbits/dylint/blob/master/utils/linting/src/lib.rs#L647-L651)

```rust
let value = config::try_init_with_metadata(&metadata)?;

if let Some(s) = &value {
    sess.file_depinfo().lock().insert(Symbol::intern(s));
}
```

`try_init_with_metadata` returns `read_to_string(&dylint_toml)` ([`internal/src/config.rs:70-76`](https://github.com/trailofbits/dylint/blob/master/internal/src/config.rs#L70-L76)) — the file contents, not its path.

The `DYLINT_TOML` env-var branch a few lines above is correct; it uses `env_depinfo`.

## Repro

Any workspace with a `dylint.toml`. With a config whose first line is a comment:

```
$ cargo check --workspace     # warm, no source changes
   Checking ...               # every workspace crate, every time
```

`CARGO_LOG=cargo::core::compiler::fingerprint=trace` shows:

```
stale: missing "/path/to/workspace/# my-lints per-lint configuration for acme."
fingerprint dirty for acme-core v0.1.0
    dirty: FsStatusOutdated(StaleItem(MissingFile { path: "/path/to/workspace/# my-lints per-lint configuration for acme." }))
```

The generated `.d` file contains the config verbatim as dependency paths:

```
acme_util-01a049eb36deb0f3.d: crates/acme-util/src/lib.rs ... #\ my-lints\ per-lint\ configuration\ for\ acme.
#
#\ acme\ is\ a\ compiler\ /\ IDE-infrastructure\ codebase.\ ...
```

Comments are not required — a config with only `[table]` headers reports `stale: missing ".../[some_lint]"`.

## Impact

Measured on two workspaces, warm runs with no source changes:

| workspace | with `dylint.toml` | config via `DYLINT_TOML` |
|---|---|---|
| 19 crates, 532k LOC | 10.1 s | 0.14 s |
| 7 crates, 47k LOC | 18.3 s | 0.68 s |

Deleting `dylint.toml` entirely gives the same speedup, confirming the file is the sole cause.

## Fix

Insert the path rather than the contents. `try_init_with_metadata` already computes `workspace_root.join("dylint.toml")`; returning that (or returning both) would let the call site record the correct value.

## Workaround

Pass the config through the environment, which takes the `env_depinfo` path:

```sh
DYLINT_TOML="$(cat dylint.toml)" cargo dylint --lib my_lints
```

Verified equivalent: identical findings (55 sites, byte-for-byte) with and without, and a config table added this way still suppresses its lint.

## Versions

`dylint_linting` 6.0.2, `cargo-dylint` 6.0.2, nightly-2026-04-16, Linux x86_64. Present on `master` as of this writing.

rirze:
Fixes #2015.

When a workspace has a `dylint.toml`, its contents are currently inserted into rustc's `file_depinfo` as a filename. Cargo therefore treats unchanged crates as dirty because that file does not exist.

Return the config file's path from `try_init_with_metadata` and load it through `sess.source_map().load_file(...)`, as suggested in the issue. This lets rustc track the actual config file and removes the unused `FileDepInfo` compatibility trait.

Extend the existing integration test to require `Fresh` results on unchanged runs, both before and after a config edit. It also verifies that editing the config changes lint diagnostics and deleting it restores the defaults. The new assertion fails before the fix with `the file [unnamed_constant] is missing` and passes afterward.

Validation on Linux, using the example's pinned `nightly-2026-07-09`:

- `cargo test -p cargo-dylint --no-default-features --features=cargo-cli --test integration depinfo_dylint_toml`
- `cargo test -p cargo-dylint --no-default-features --features=cargo-cli --test integration library_packages_in_dylint_toml`
- `cargo clippy -p cargo-dylint --no-default-features --features=cargo-cli --test integration -- -D warnings`
- `cargo fmt --all -- --check`, plus nightly rustfmt on all three changed files.


smoelius:
Thanks, @rirze! I will try to review this within the next few days.

internal/src/config.rs
```diff
@@ -63,26 +63,22 @@ pub fn try_init_with_metadata(metadata: &cargo_metadata::Metadata) -> Result<Opt
...
     }
 
-    Ok(value)
+    let value = read_to_string(&dylint_toml).map_err(|error| {
+        Inner::Io(
+            format!("`read_to_string` failed for {dylint_toml:?}"),
+            error,
+        )
+    })?;
+    init_from_string(&value)?;
```
smoelius, line 80:
```suggestion

    init_from_string(&value)?;
```
Nit.

utils/linting/src/lib.rs
```diff
@@ -734,10 +705,10 @@ fn try_init_config_guarded(sess: &rustc_session::Session) -> ConfigResult<()> {
         _ => {
             let metadata = result?;
 
-            let value = config::try_init_with_metadata(&metadata)?;
-
-            if let Some(s) = &value {
-                sess.file_depinfo().lock().insert(Symbol::intern(s));
+            if let Some(path) = config::try_init_with_metadata(&metadata)? {
+                sess.source_map().load_file(&path).map_err(|error| {
+                    ConfigError::other(format!("`load_file` failed for {path:?}: {error}"))
```
smoelius:
```suggestion
                    ConfigError::other(format!("Could not load {path:?}: {error}"))
```
I know both forms of error message are used ("\<fn\> failed for" and "Could not \<verb\>"), and not necessarily consistently.

But since `load_file` is a rustc API and could change, it is probably better to use the "Could not \<verb\>" form.

smoelius (commented):
Thanks very much for this, @rirze. Two very minor nits, and then I will merge.

Aside: you PR revealed another bug (#2079), which is that creating a dylint.toml file when one did not already exist does not cause the lints to be rerun. You should not feel obligated to fix this, however.

smoelius:
Also, if you agree with my suggestions, please squash your changes into one commit.

rirze:
Thanks for getting back to me-- I made the changes you noted and squashed the commits into one. Awaiting checks to complete and then I'll leave this PR in your hands.

Regarding #2079 , I think I can submit a PR for that shortly today.

PR #2077, opened 2026-09-13
Add --fail-on-no-libraries option to cargo dylint

### LINKED ISSUE #2073: cargo dylint --all exits 0 when it resolves no libraries
## What happens

`cargo dylint --all` exits 0 when it resolves no libraries. It prints only `Warning: No libraries were found.` and checks nothing.

In CI that reads as success. A workspace whose `[workspace.metadata.dylint]` section is missing, renamed or moved keeps a green lint gate, and `cargo clippy` does not catch it either.

Measured on cargo-dylint 6.0.4. A `pattern` that matches nothing, and an unreachable `rev`, both fail properly. Only the empty set is silent.

## Where

`warn_if_empty` (`src/lib.rs:145-154`) reports the empty set through a return value the caller uses only for an `assert!` (`:124-135`). Nothing returns early, so `check_or_fix` iterates an empty map (`:318`) and returns `Ok(())`.

## Ask

An opt-in flag would be enough, say `--fail-on-no-libraries`. Our workaround is to grep `cargo dylint list` before the run.

humb1t:
Closes #2073.

## Problem

`cargo dylint --all` exits 0 when it resolves no libraries, printing only `Warning: No libraries were found.`. In CI that reads as success, so a workspace whose `[workspace.metadata.dylint]` section is missing, renamed, or moved keeps a green lint gate.

## Solution

Add an opt-in `--fail-on-no-libraries` flag. When a check resolves zero libraries, `cargo dylint` now exits with an error instead of silently succeeding:

```
$ cargo dylint --all --fail-on-no-libraries
Warning: No libraries were found.
Error: No libraries were found.
```

Scope:

- The flag applies to the check operation only; `cargo dylint list` keeps the warning-only behavior.
- The "Nothing to do. Did you forget `--all`?" path is unaffected (usage hint, not empty resolution).
- Default behavior is unchanged.

## Testing

- New integration test `warn::fail_on_no_libraries` covers: failure with `--all` in an empty workspace, `list` unaffected, and flag-as-no-op without a library selection.
- Existing `warn` tests pin the unchanged default behavior.
- `cargo fmt --check` and `cargo clippy -p dylint -p cargo-dylint --all-targets` are clean.

Assisted-by: GLM 5.3 Flash & Qwen 3.8 (via opencode)


smoelius:
@humb1t Thanks very much for this.

Bottom line: Do you mind if I push to your branch?

There are several small changes that I would like to make. Some of them may require changes to existing code that you should not have to make. So, out of respect for your time, I would like to make them myself. I will give you a chance to review before I merge.

Below are some of the high points.

I would like the handling of the default check and list operations to be more uniform. So, for example, I would prefer that this command fail rather than succeed:
```rust
    cargo_dylint()
        .current_dir("../fixtures/empty")
        .args(["dylint", "list", "--all", "--fail-on-no-libraries"])
        .assert()
        .success()
        .stderr(predicate::str::ends_with(
            "Warning: No libraries were found.\n",
        ));
```
Also, I would like to make the help message more explicit by saying:
```
"Exit with an error if no libraries are found; requires a library selection \
 (`--all`, `--git`, `--lib`, `--lib-path`, or `--path`)"
```
But taking that literally implies that this command should also fail:
```rust
    cargo_dylint()
        .current_dir("../fixtures/empty")
        .args(["dylint", "--fail-on-no-libraries"])
        .assert()
        .success()
        .stderr(predicate::str::ends_with(
            "Warning: Nothing to do. Did you forget `--all`?\n",
        ));
```
And by the same token, `cargo dylint list --fail-on-no-libraries` should fail as well, though there's no test for that case today.

So I hope you won't mind if I push to your branch. But please let me know if you would.

humb1t:
You can push for sure! I don't have any design preference. As an author I think you better understand what is the "right" way, so please 🚀

smoelius:
@humb1t Could you please review my changes in https://github.com/trailofbits/dylint/pull/2077/changes/e3be24e2d6b2289e4caa3205b21bb0569f104047? In particular, please tell me if any of them would make `--fail-on-no-libraries` no longer suitable for your needs.

I think Claude's commit message is a pretty good description of the changes.

humb1t:
@smoelius - thank you a lot, tests showed that your changes are superior to mine, so big thanks for improving this PR:

| case | 6.0.4 | e3be24e |
|---|---|---|
| `dylint --all` | exit 0, `Warning: No libraries were found.` | exit 0 (default unchanged) |
| `dylint --all --fail-on-no-libraries` | — | exit 1, `Error: No libraries were found`, no warning |
| `dylint --path <dir with no library>` | exit 0 | — |
| `dylint --path <dir with no library> --fail-on-no-libraries` | — | exit 1 |
| `dylint list --all --fail-on-no-libraries` | — | exit 1 |
| `--fail-on-no-libraries` with no selection (check or `list`) | — | exit 1, names the flags |
| `dylint --lib-path <real .so> --fail-on-no-libraries` | — | exit 0, checks normally |

From my side PR is ready to be merged 🚀

smoelius:
Thanks, @humb1t!

PR #2083, opened 2026-09-14
Invalidate lints when dylint.toml is created

### LINKED ISSUE #2079: Creating a `dylint.toml` does not cause lints to rerun
*Filed by Claude Code on behalf of @smoelius.*

## Summary

When a workspace has no `dylint.toml`, `try_init_with_metadata` registers nothing as a compilation
input. Creating the file afterward does not invalidate Cargo's fingerprint, so Cargo reports the
unit as `Fresh`, the driver never runs, and lints keep using their default configuration.

This is pre-existing and does not require #2072. It is currently masked: an existing `dylint.toml`
puts a nonexistent filename into the dep-info, so every unit is dirty on every run and the config is
always reread. #2072 fixes that, which makes this case observable.

## Scope

This affects configuration only. If the new `dylint.toml` adds a `[[libraries]]` entry, the library
list changes, `DYLINT_LIBS` changes with it, and `sess.env_depinfo` causes Cargo to re-invoke the
driver. A file containing only lint configuration — `[unnamed_constant] threshold = 1` — changes
none of the tracked variables.

## Reproduction

Against `master`, in a workspace whose `Cargo.toml` names
`examples/supplementary/unnamed_constant` as a library, with a `src/main.rs` containing a `9`
literal:

```console
$ cargo dylint --all -- --verbose
    Checking absent_present_test v0.1.0 (...)
# no warning: 9 is below the default threshold of 10

$ printf '[unnamed_constant]\nthreshold = 1\n' > dylint.toml

$ cargo dylint --all -- --verbose
       Fresh absent_present_test v0.1.0 (...)
# no warning: the driver never ran, so the new threshold was never read

$ touch src/main.rs && cargo dylint --all -- --verbose
    Checking absent_present_test v0.1.0 (...)
warning: unnamed constant
```

The third command is the control. The configuration is valid and does take effect once anything
else forces a rebuild, which confirms the second run is a staleness problem rather than a problem
reading the file.

## Why the obvious fix does not work

Registering the path unconditionally — returning `workspace_root.join("dylint.toml")` whether or not
it exists — fails either way it is tracked:

- Through `sess.source_map().load_file`, which returns `Err` for a nonexistent file. Every workspace
  without a `dylint.toml` would fail to compile.
- Through `file_depinfo`, which puts the path in the `.d` file. Cargo cannot stat it, so the unit is
  dirty on every run. That is the bug #2072 fixes, reintroduced for the no-config case.

## Proposed direction

Cargo's fingerprint can be reached with a value rather than a file. `cargo-dylint` runs on every
invocation regardless of freshness, so it can compute a digest of `dylint.toml` — with a distinct
value for "absent" — and set it alongside `DYLINT_LIBS` and `DYLINT_METADATA` in
`dylint/src/lib.rs`. Adding it to `UNTRACKED_STATE_VARS` would write it into `sess.env_depinfo`, and
Cargo would compare the recorded value against the environment on the next run.

The digest has to come from `cargo-dylint`, not the driver: the driver runs only after Cargo has
already decided the unit is stale, which is the state this is meant to detect.

Whether the absent-to-present transition additionally requires rustc to discard its incremental
cache is untested. `UNTRACKED_STATE_VARS` feeds both consumers, but the rustc-side one currently has
no working mechanism — see #2078.

## References

- #2072 — fixes dependency tracking for an existing `dylint.toml`
- #2015 — the bug #2072 fixes
- #2078 — `--env-set` removal

---

rirze:
Fixes #2079.

`cargo-dylint` now computes an absent-or-content digest for the workspace `dylint.toml` and passes it to the driver. The driver tracks that value in Cargo dep-info and Rustc’s incremental state, so creating or changing the configuration reruns configured lints.

A dedicated `DYLINT_TOML_DIGEST` variable keeps the existing environment contracts intact. `DYLINT_TOML` is an explicit TOML-content override consumed by the linting crate, while `DYLINT_METADATA` is serialized workspace library metadata. The driver cannot compute the digest itself because Cargo invokes it only after deciding whether a crate is fresh.

The integration test starts without `dylint.toml`, creates one that lowers `unnamed_constant`’s threshold, and verifies that Cargo checks the crate again and emits the lint.

Validation:

- `cargo test -p cargo-dylint --no-default-features --features=cargo-cli --test integration depinfo_dylint_toml` (with an isolated freshly built driver)
- `cargo clippy -p dylint --features library_packages -- -D warnings`
- `cargo +nightly-2026-07-09 clippy --manifest-path driver/Cargo.toml -- -D warnings`
- `cargo +nightly-2026-07-09 fmt --all -- --check`


cargo-dylint/tests/integration/depinfo_dylint_toml.rs
```diff
@@ -60,6 +60,23 @@ path = "{}/../examples/supplementary/unnamed_constant"
 
     let dylint_toml = tempdir.path().join("dylint.toml");
 
+    cargo_bin_cmd!("cargo-dylint")
+        .current_dir(&tempdir)
+        .args(["dylint", "--all", "--", "--verbose"])
+        .assert()
+        .success()
+        .stderr(predicate::str::contains("warning: unnamed constant").not());
```
smoelius:
```suggestion
        .stderr(
            predicate::str::contains("Checking depinfo_dylint_toml_test")
                .and(predicate::str::contains("warning: unnamed constant").not()),
        );
```

cargo-dylint/tests/integration/depinfo_dylint_toml.rs
```diff
@@ -60,6 +60,23 @@ path = "{}/../examples/supplementary/unnamed_constant"
 
     let dylint_toml = tempdir.path().join("dylint.toml");
```
smoelius: Please put your test additions above this line, so that the path calculation stays next to its first use.

In other words, following your additions, the test should look like:
```rust
    ...

    cargo_bin_cmd!("cargo-dylint")... /// <-- your first added statment

    cargo_bin_cmd!("cargo-dylint")... /// <-- your second added statment

    let dylint_toml = tempdir.path().join("dylint.toml");

    write(&dylint_toml, "[unnamed_constant]\nthreshold = 100\n").unwrap();

    ...
```

cargo-dylint/tests/integration/depinfo_dylint_toml.rs
smoelius: I think this should work. It makes the check more precise (and follows the pattern of a check near the end of this test).

dylint/src/lib.rs
```diff
@@ -474,6 +477,32 @@ fn clippy_disable_docs_links() -> Result<String> {
     serde_json::to_string(&val).map_err(Into::into)
 }
 
+/// Returns a value that changes when the workspace's `dylint.toml` changes or is created.
+fn dylint_toml_digest(opts: &opts::Dylint) -> Result<String> {
+    let mut command = MetadataCommand::new();
+    if let Some(path) = &opts.library_selection().manifest_path {
+        command.manifest_path(path);
+    }
+    let metadata = command.no_deps().exec()?;
```
smoelius: These same lines begin `target_dir` below.

Could you please extract them into helper function like the following, place it just below `target_dir`, and make both `dylint_toml_digest` and `target_dir` call it?
```rust
fn metadata_from_library_selection(opts: &opts::Dylint) -> Result<Metadata> {
    let mut command = MetadataCommand::new();
    if let Some(path) = &opts.library_selection().manifest_path {
        command.manifest_path(path);
    }
    command.no_deps().exec().map_err(Into::into)
}
```

dylint/src/lib.rs
```diff
@@ -474,6 +477,32 @@ fn clippy_disable_docs_links() -> Result<String> {
...
+    let dylint_toml = metadata.workspace_root.join("dylint.toml");
+
+    let mut hasher = DefaultHasher::new();
+    hasher.write(b"dylint-toml-digest-v1");
+    let exists = dylint_toml
+        .try_exists()
+        .with_context(|| format!("Could not determine whether {dylint_toml:?} exists"))?;
+    if exists {
+        hasher.write(b"present");
```
smoelius:
```suggestion
        hasher.write(b"\x07present");
```
Writing header bytes followed by content bytes with nothing to distinguish them feels like it could bite us down the road. The suggestion to use `0x07` (i.e., the length of "present") is based on Ethereum's [EIP-712](https://eips.ethereum.org/EIPS/eip-712).

dylint/src/lib.rs
```diff
@@ -474,6 +477,32 @@ fn clippy_disable_docs_links() -> Result<String> {
...
+        let contents =
+            read(&dylint_toml).with_context(|| format!("Could not read {dylint_toml:?}"))?;
+        hasher.write(&contents);
+    } else {
+        hasher.write(b"absent");
```
smoelius:
```suggestion
        hasher.write(b"\x06absent");
```
Technically, "absent" is not followed by content bytes, but this follows the pattern established by "\x07present".

smoelius (commented):
@rirze Thanks very much for this, and sorry it has taken me so long to reply.

Hopefully, each of my comments is easy to address. If you could, please squash all of your commits into one.

Also, please rebase onto the current `master`.

smoelius:
Thanks, @rirze!

PR #2097, opened 2026-09-18
Honor RUSTC when detecting nightly in dylint_internal

dzbarsky:
Use `RUSTC` when checking whether the compiler supports nightly options. Build system can select a compiler that differs from `rustc` on `PATH`; checking the latter sets the wrong `nightly` configuration. Keep the existing `rustc` fallback when `RUSTC` is unset.

Validated the standalone build script with selected nightly and stable compiler stubs and with `RUSTC` unset. The selected-nightly case fails before this change and passes afterward.

Assisted by OpenAI Codex.


smoelius:
Thanks for your PR, @dzbarsky. But could you expand a little on how you ran into this?

> Build system can select a compiler that differs from `rustc` on `PATH`; checking the latter sets the wrong `nightly` configuration.

I would like to better understand the problem before deciding on a solution.

dzbarsky:
> Thanks for your PR, @dzbarsky. But could you expand a little on how you ran into this?
> 
> > Build system can select a compiler that differs from `rustc` on `PATH`; checking the latter sets the wrong `nightly` configuration.
> 
> I would like to better understand the problem before deciding on a solution.

We use Bazel, not Cargo. Bazel provides hermetic paths to all actions and does not rely on PATH lookup, it also sandboxes the actions and can ship the entire set of inputs to a remote executor on a per-action basis.

smoelius:
Is this the only way that lack of support for `RUSTC` affects you?

That is, does this patch completely resolve your problems regarding `RUSTC`?

dzbarsky:
Yes with this fixed we have been using dylint nicely for months and finding a lot of value in it! 💖

smoelius:
Thank you for the feedback, which I very much appreciate. 🙏

I think this patch looks good.

I will try to get a release with it out soon, hopefully within the next few days.

